PR #6, opened 2020-12-17
add tests for examples

MarcoGorelli:
moving these over from pymc3


MarcoGorelli:
kept in PyMC3 in the end

PR #7, opened 2020-12-18
Fixup dependent density regression

MarcoGorelli:
'Dependent density regression notebook' currently doesn't run (the data doesn't download, and the plot of the posterior predictive throws an error) so here's a PR to fix it up.

Furthermore, the notebook previously read

> Since only three mixture components have appreciable posterior expected weight for any data point, we can be fairly certain that truncation did not unduly influence our results.

yet the posterior plot only showed _two_ mixture components having appreciable posterior expected weight for any data point. With the current re-run, it looks right


AlexAndorra:
Is this ready for review?

MarcoGorelli:
Sure!

MarcoGorelli:
@AlexAndorra ready now, thanks!

examples/mixture_models/dependent_density_regression.ipynb
```diff
@@ -2,7 +2,16 @@
  "cells": [
```
AlexAndorra, line 2: <p>This is fine, but I think you could do the handling on <code>w</code> directly with <code>xarray</code>:  http://xarray.pydata.org/en/stable/computation.html<span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/7/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

AlexAndorra (approved):
This looks nice, thanks @MarcoGorelli !

examples/mixture_models/dependent_density_regression.ipynb
MarcoGorelli: ah, thanks! OK I'll do that next time

PR #10, opened 2020-12-19
remove old location of notebook from examples

MarcoGorelli:
`GLM-hierarchical-advi-minibatch` appears twice in the table of contents, both in
```
generalized_linear_models/GLM-hierarchical-advi-minibatch
```
and in 
```
variational_inference/GLM-hierarchical-advi-minibatch
```

The latter is correct.

Eventually, there should probably be a CI job in PyMC3 which builds the docs and would catch this kind of error (which prevents the docs from being built)


AlexAndorra (approved):
Thanks @MarcoGorelli !

PR #18, opened 2021-01-04
Dirichlet mixture of multinomials example

bsmith89:
Add an example to demonstrate the forthcoming DM distribution (see [PyMC3 PR #4373](https://github.com/pymc-devs/pymc3/pull/4373)).


AlexAndorra:
Ooh nice @bsmith89 ! I'll review by tomorrow night 🥳

bsmith89:
> Any reason you couldn't use fast_sample_posterior_predictive instead of sample_posterior_predictive?

Only because I didn't know that function existed until right now 🙃 

> Also, can you put the predictive samples into the InferenceData ?

Can you point me to docs on how to modify an InferenceData object?

> Could you plot a comparison between the predicted and observed samples?

Input/Output cell 12 is a very superficial version of this already, no. Do you have ideas about deeper comparisons to plot?

rpgoldman:
> > Also, can you put the predictive samples into the InferenceData ?
> 
> Can you point me to docs on how to modify an InferenceData object?

I believe there are two ways you could do this:

1. Create a new `InferenceData` object, with `az.from_pymc3()`, and this time adding the keyword argument `posterior_predictive` with the return value from (`fast_`)`sample_posterior_predictive`.
2. Use `az.from_pymc3_predictions`, but I don't think this is actually necessary, since you are doing predictive sampling with the same independent variable values and shapes.

LMK if that won't work, and I will try to help. Seems like there ought to be a way of creating a new `InferenceData` with only the PP samples and merge it into the original one, but I don't know how to do that.

> > Could you plot a comparison between the predicted and observed samples?
> 
> Input/Output cell 12 is a very superficial version of this already, no? Do you have ideas about deeper comparisons to plot?

I'm sorry, I don't.  I was hoping you did!  It seems like such a comparison is something that ArviZ should help with, but I don't know of an ArviZ plot function to use.

bsmith89:
> LMK if that won't work, and I will try to help. Seems like there ought to be a way of creating a new InferenceData with only the PP samples and merge it into the original one, but I don't know how to do that.

Turns out there is!

```python
with model_marginalized:
    ppc = pm.fast_sample_posterior_predictive(trace_marginalized, keep_size=True)
    
# Concatenate with InferenceData object
trace_marginalized = az.concat(trace_marginalized, az.from_dict(posterior_predictive=ppc))
```

> Maybe arviz.plot_ppc()?

Messed around with this a bit just now, and I can't really get much that seems worthwhile. For now I think I'll stick with the superficial example, but using the more complete InferenceData object as constructed above.

rpgoldman:
> > Maybe arviz.plot_ppc()?
> 
> Messed around with this a bit just now, and I can't really get much that seems worthwhile. For now I think I'll stick with the superficial example, but using the more complete InferenceData object as constructed above.

Works for me.  I'm not sure that ArviZ has the right capabilities for such PP checks, but it's likely I just don't know all that's in there.

bsmith89:
> Works for me. I'm not sure that ArviZ has the right capabilities for such PP checks, but it's likely I just don't know all that's in there.

`pm.plot_ppc` may well be the right function for many PPCs, but this particular example is challenging because it's multivariate. 🤷‍♂️

rpgoldman:
Maybe put an issue on the ArviZ GitHub issues page to point out this limitation?

AlexAndorra:
I'm not sure yet it's a limitation from ArviZ actually (see my comment for this cell on the NB). I'd bet this is doable with the `coords` kwarg, which would be neat because you're already returning an `InferenceData`object 😉

AlexAndorra (changes requested):
This is a really good start @bsmith89 🤩 
Overall, I think this notebook lacks context about the problem and model. I left comments below, trying to adopt the point of view of a user who thinks this model is what she needs but is not sure yet -- feel free to ask if anything is unclear.
Once these comments are adressed it should be ready to merge 🎉

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
```diff
@@ -0,0 +1,748 @@
+{
```
AlexAndorra, line 1: <p>I think you should introduce the example, data and parameters -- it'll ease understanding for readers: what's this type of model? In which contexts can it be useful? What are its strenghts and weaknesses?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

AlexAndorra, line 1: <p>Same thing here: </p><ul><li>What's the model doing? </li><li>What do the parameters mean? </li><li>How are you choosing the priors?</li><li>Why are you parametrizing the Dirichlet with <code>a * alpha</code> instead of the classical <code>alpha</code> ? </li><li>Are you expecting this model to run well?</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

AlexAndorra, line 1: <p>Add a trailing <code>;</code> at the end to get rid of MPL axes objects printing</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

AlexAndorra, line 1: <p>How satisfied are you with these diagnostics and trace plot? How does it motivate the use of the coming mixture model?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

AlexAndorra, line 1: <p>How does this model solve the problem you were having above?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

AlexAndorra, line 1: <p>Add trailing <code>;</code>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

AlexAndorra, line 1: <p>This should quickly be put into context too: why are you doing this? Why is it useful?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

AlexAndorra, line 1: <p>I'm positive you can use <code>az.plot_ppc</code> with the <code>kind</code> and <code>coords</code> kwargs here (see https://arviz-devs.github.io/arviz/api/generated/arviz.plot_ppc.html).</p><p>The nice is that you'll use <code>xarray</code> indexing instead of classic numpy indexing. <code>i</code> and <code>j</code> are the sample and category, right?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

AlexAndorra:
Hey @bsmith89 ! Did you get some time to make the changes? Any questions?
As the PR on the PyMC repo is close to merging, it'd be great to add this tutorial simultaneously and have all this in the next release 🎉

bsmith89:
I have not made changes yet, no. @ricardoV94 and I were discussing potentially making a more involved topic modeling example. You can see some of that discussion in this [Discourse thread](https://discourse.pymc.io/t/implementing-the-dirichletmultinomial-distribution-in-pymc3/6472/40). It's been challenging to find a realistic example without running into problems during sampling.

If the goal is to get this merged before the next release I think making the changes you have suggested to this existing example is probably the best approach. I'll take another look this evening or tomorrow.

AlexAndorra:
I see, thanks!

>  @ricardoV94 and I were discussing potentially making a more involved topic modeling example

Well, we definitely want polished and well tought-out tutorial NBs, so that sounds awesome! And let's not rush it -- quality is more important than speed here. How much time do you guys think you need to write down this kind of example?

Depending on this, we could merge the current simple tutorial (after making the changes I have suggested) to get something out with the next release, and replace it with the full case study when the latter is out?

bsmith89:
> How much time do you guys think you need to write down this kind of example?

Yes, I think it will be challenging to identify a state-of-the-art topic-modeling example that doesn't run into sampling challenges (because most models are symmetrical to permutations of the topic labels, and therefore multimodal). Clearly we have dealt with that in other examples (e.g. [the matrix factorization example](https://docs.pymc.io/notebooks/probabilistic_matrix_factorization.html), but I think it will require some thought about the best thing to do for our audience.

> Depending on this, we could merge the current simple tutorial (after making the changes I have suggested) to get something out with the next release, and replace it with the full case study when the latter is out?

This is my vote. I think, with the changes you suggested, the current example will get users up-and-running. We can then extend it with a more realistic example of e.g. topic modeling after that.

AlexAndorra:
> I think it will require some thought about the best thing to do for our audience.

That makes total sense. Really looking forward to reviewing this -- hope I can be of help! And feel free to open questions on Discourse about this 😉

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
bsmith89: I can get `plot_ppc` to plot something using the `coords=` kwarg but it's not at all clear to me that it's set up well to produce a histogram like the one I made manually.

Here's what I tried:
```python
az.plot_ppc(
    trace_marginalized,
    kind='scatter',
    var_names=['obs'],
    coords=dict(obs_dim_0=[9], obs_dim_1=[2])
)
```

And here's the result:
image

Having not read up on PPCs in ArviZ, it's not super apparent to me what that plot's trying to show. What's the y-axis? Frequency I assume, but then what are the `'o'`-styled points? In any case, I'd much prefer to use a bar-chart style histogram than a KDE-plot, given that the values of `obs` are discrete integers.

Anyway, that's all to say that I don't think `plot_ppc` is currently the correct tool for this job.

Instead, to motivate the use of a PPC, I think I'll run a _third_ model that's just a plain Multinomial distribution, and use the PPC to demonstrate that it does not fit the over-dispersed data well.

AlexAndorra: Interesting 🤔 Let me take a look ASAP, because I'm still not convinced that it's not currently possible with `plot_ppc`. Did you allow push by maintainers on this PR?
Also tagging @OriolAbril and @aloctavodia, who are ArviZ wizards 🧙‍♂️ 

>  I'll run a third model that's just a plain Multinomial distribution, and use the PPC to demonstrate that it does not fit the over-dispersed data well.

Great idea!

bsmith89: > I'm still not convinced that it's not currently possible with plot_ppc.

Looking forward to being proven wrong 😄

> Did you allow push by maintainers on this PR?

Looks like that box is checked, yes.

AlexAndorra: Perfect, thanks. I'll tell you when I got some time to look over it

> Looking forward to being proven wrong 😄

Ha ha, sorry, that's not what I meant -- rather, I'm seeing this as a nice, non trivial test case for `plot_ppc`'s capacities 😉

bsmith89:
Hey, sorry for the delay on this.

I've updated the notebook with what I think might be the final form of the simulation/analysis for this first iteration of the example. This includes a "null", multinomial model of the data, PPCs comparing this to the marginalized model, and a final model comparison (LOO).

What I _haven't_ done is add any of the written explanations for the code. Instead I copied your comments, @AlexAndorra, into markdown cells as placeholders for the text that should be there.

I think I can work on this for a small amount of time this week, hopefully getting it into merge-able form by Friday. However, if anyone (e.g. @ricardoV94) else would like to take a crack at writing the text, I'd be happy to relinquish that work. ;)

I'm also happy to swap out the custom PPC plotting code for an ArviZ approach, if anyone has ideas for an effective figure.

Let me know if you see more that should be changed besides the markdown cells.

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
```diff
@@ -0,0 +1,1360 @@
...
```
ricardoV94, line 1: <p>A model description (Latex and Graphviz) would be great</p><p> </p><pre class="ql-syntax" spellcheck="false">p[0,...,9] ~ Dirichlet(alpha)
y[0,...,9] ~ Multinomial(n, p[0,...,9])
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

ricardoV94, line 1: <p>Looks like a scipy issue in the multinomial. This seems to run without problems on my machine:</p><p> </p><p><code>p = sp.stats.dirichlet(conc * frac).rvs(size=n_samples)</code></p><p><code>y = np.stack([np.random.multinomial(20, p[i, :]) for i in range(n_samples)])</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

ricardoV94, line 1: <p>I would make the analogy to the BetaBinomial. The same way that the BB can be used to model overdispersed Binomial data (by assuming each batch of data follows a slightly different p, from a common Beta), the DM can be used to model overdispersed Multinomial data (by assuming each batch of data follows a slightly different p, from a common Dirichlet).</p><p> </p><p>One thing that has confused me in the past is that this "different" p applies for a batch of size n, and not for every trial observed. I used to think it was the latter and this was confusing for me. Not sure if others feel the same way.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

ricardoV94:
I left some comments above. I think the NB is very interesting. It would be cool to have a clear posterior predictive sampling plot that shows how the Multinomial-only model generates underdispersed predictions relative to the observed data.

I would add graphviz models to the 3 models, but that is a personal preference that others may not share.

bsmith89:
I think I made all the changes we talked about, and this is ready for final review.

Let me know what you think!

ricardoV94:
Looks really nice. I will have a couple of minor suggestions when I am on the desktop, but I think it is already great!

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
MarcoGorelli: <p>Or even</p><pre class="ql-syntax" spellcheck="false">np.vstack([sp.stats.multinomial(n=total_count, p=p[i, :]).rvs() for i in range(n_samples)])
</pre><p>works for me</p>

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
```diff
@@ -0,0 +1,1819 @@
...
```
MarcoGorelli, line 1: <p>bit confused about what his <code>None</code> is, in the second line of this cell</p><p> </p><p>EDIT: just saw Alex's comments on this, agree that <code>;</code>  would be better</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

MarcoGorelli, line 1: <p>97% HDIs?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

MarcoGorelli, line 1: <p>And now (instead of "And no")</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

MarcoGorelli:
Nice work, this was interesting to read, thanks!

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
ricardoV94: <p>Seconded</p>

ricardoV94: <p>Not 94%?</p>

ricardoV94, line 1: <blockquote>the DM has one additional parameter that controls this</blockquote><p>It's not strickly one additional parameter right? It uses a different (random) parameter. Maybe something like: unlike the Multinomial where the probability vector is fixed, the DM allows for a slightly different probability vector for each batch of observations, which is drawn from a common Dirichlet distribution.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

ricardoV94, line 1: <p>conc: a concentration parameter that controls the how much the probability vectors for each batch of observations can vary. Low concentration values allow for high variance (over dispersed data), while larger values lead to more similarly distributed data. In the limit of a large concentration, the model approaches the Multinomial distribution. (Not the exact wording, just a suggestion).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

ricardoV94, line 1: <p>Why not iterate directly over the individual p rows instead of over the indexes?</p><p> </p><pre class="ql-syntax" spellcheck="false">y = np.stack([sp.stats.multinomial(n=total_count, p=p_vector).rvs()[0] for p_vector in p])
</pre><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

ricardoV94, line 1: <p>Null multinomial model -&gt; Vanilla / Plain multinomial model?</p><p> </p><p>Null has a certain connotation in the stats world</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

ricardoV94, line 1: <p>by exploding our n observations -&gt; by reparametrizing our ...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

ricardoV94, line 1: <p>This plot is really cool. I have a couple of suggestions:</p><ol><li>I wouldn't share the y-axis or clip it so that the distributions can be better seen</li><li>Maybe do clip the x-axis on the right (45 or something)</li><li>Show the true fraction (with a dashed vertical line or something not very overwhelming). This way you clearly show not only the bias in variance but also the bias in location.</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

ricardoV94, line 1: <p>First time you use the word taxon? Maybe just stick with species for the non-biologist readers</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

ricardoV94, line 1: <p>Dirichlet-Multinomial Model</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

ricardoV94, line 1: <p>Maybe not mention this if you are not going to explore where they are occurring...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

ricardoV94, line 1: <p>Marginalized Dirichlet-Multinomial model</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

ricardoV94, line 1: <p>but now marginalizing over the latent <code>pi</code>  dirichlet parameters</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

ricardoV94, line 1: <p>out-of-sample ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

ricardoV94, line 1: <p>Besides the Generalized Dirichlet you can also mention the Logistic Multivariate Normal distribution: <a href="https://en.wikipedia.org/wiki/Logit-normal_distribution#Multivariate_generalization" target="_blank">https://en.wikipedia.org/wiki/Logit-normal_distribution#Multivariate_generalization</a></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='47'/>

ricardoV94, line 1: <p>typo: beliefs -&gt; belief</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

ricardoV94:
@bsmith89 I left some suggestions. Let me know if you have any questions :)

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
MarcoGorelli: <p>or</p><pre class="ql-syntax" spellcheck="false">np.vstack([sp.stats.multinomial(n=total_count, p=p_vector).rvs() for p_vector in p])
</pre><p>so there's no need for the mysterious <code>[0]</code>?</p>

bsmith89: <blockquote>It's not strickly one additional parameter right? It uses a different (random) parameter.</blockquote><p> </p><p>Hmm...I'm not sure that distinction makes sense to me. The multinomial is parameterized by p on the k-simplex (k-1 dimensions), and the DM is parameterized by alpha in the positive reals (k dimensions). Am I using "parameter" differently?</p><p> </p><blockquote>unlike the Multinomial where the probability vector is fixed, the DM allows for a slightly different probability vector for each batch of observations, which is drawn from a common Dirichlet distribution.</blockquote><p> </p><p>I'm fine with using this intuitive explanation for what the DM is, but it may be redundant with some of the explanation that's currently below, closer to the simulation (i.e. "Each observation from the DM is simulated by..."). I can imagine an argument for including this information near the top, though.</p><p> </p>

bsmith89: <blockquote>a concentration parameter that controls the how much the probability vectors for each batch of observations can vary. Low concentration values allow for high variance (over dispersed data), while larger values lead to more similarly distributed data. In the limit of a large concentration, the model approaches the Multinomial distribution</blockquote><p> </p><p>I feel like I already have most of that, but I could be missing something. Can you explain more about which details in your suggested text are missing or less clear in the current description of the <code>conc</code> parameter:</p><p> </p><blockquote>controlling the overdispersion of our data, where larger values result in our distribution more closely approximating the multinomial. </blockquote><p> </p><p>Is "overdispersion" to lingo-y and you're suggesting that I focus on referring to "variance"?</p>

bsmith89: <blockquote>Why not iterate directly over the individual p rows instead of over the indexes?</blockquote><p> </p><p>Yep, good call, @ricardoV94. Definitely more pythonic.</p><p> </p><pre class="ql-syntax" spellcheck="false">np.vstack([sp.stats.multinomial(n=total_count, p=p_vector).rvs() for p_vector in p])
</pre><p> </p><p>Ooh, I like that, @MarcoGorelli. Will use.</p>

bsmith89: <blockquote>Null has a certain connotation in the stats world</blockquote><p> </p><p>Yeah, I was torn about some of the model "names" (along with the naming of other python variables, too).</p><p> </p><p>I'm down for <code>model_multinomial</code> , <code>model_dm_explicit</code>,  and <code>model_dm_marginalized</code>. Do those seem good to you?</p>

bsmith89: <p>I thought "exploding" was more evocative and pairs well with "collapsing". How do you feel about:</p><p> </p><blockquote>Because of a couple of identities of the multinomial distribution, we can reparameterize this model in a number of ways—we could explode...</blockquote>

bsmith89: <p>Yeah, that seems to be most folks' preference... :-/</p><p> </p><p>I'll make that change for consistency with the other examples.</p><p> </p><p>(FWIW: My arguments for <code>None</code> is that  the <code>;</code>  at the end of a line is (A) hidden and (B) can be confusing for python beginners who don't really see semicolons anywhere else in the language. I'd also guess that (C) with enough copy-pasting semicolons will start polluting other lines.)</p>

bsmith89: <p>...Math is apparently not my strong-suit.</p>

bsmith89: <blockquote>I wouldn't share the y-axis or clip it so that the distributions can be better seen</blockquote><p> </p><p>Yeah, I get that. I <em>do</em> want them to be comparable across indexes, though, so maybe I could clip all of them (including the fifth) to 0.5?</p><p> </p><p>Or clip the first four there, and keep the last one at the full span?</p><p> </p><blockquote>Maybe do clip the x-axis on the right (45 or something)</blockquote><p> </p><p>I prefer the full span. It does mean there's more empty space, but I think it better communicates what the domain of the (implicit) mass function is.</p><p> </p><blockquote>Show the true fraction (with a dashed vertical line or something not very overwhelming). This way you clearly show not only the bias in variance but also the bias in location.</blockquote><p> </p><p>Great idea!</p>

bsmith89: <p>You're suggesting that I shouldn't use the latex $\pi$ when the code has the word <code>pi</code>? </p><p> </p><p>Presumably I should be consistent with this everywhere, so what are you suggesting as the overall rule for when to use which?</p><p> </p><p>Feels like there some value in a semantic separation between the <em>code</em> and the model itself, although I agree that I don't currently make that explicit in any of the text.</p>

bsmith89: <p>Good catch. I think I'll go with</p><p> </p><blockquote>If we have independent beliefs...</blockquote>

ricardoV94: Yeah I agree you say it lower. I would vote to have it only at the top

ricardoV94: I mostly meant to explicit call it concentration and explain it as such (as some might not get it from the abbreviated conc or the python code) because it is a very common type of parameter in Bayesian modeling and it might help readers connect with previous knowledge

ricardoV94: Sounds good. You can also drop the explicit and just call it dm or call it latent dm

ricardoV94: Sounded a bit strange to me. Fun verb for sure!

bsmith89: <blockquote>(Left as an exercise for the reader)</blockquote><p>?</p><p> </p><p>I'd like to keep this line because, while I'm skipping the diagnosing step and going right to the marginalized model, these divergences are still indicating a pathological geometry. I want to point that out.</p>

ricardoV94: Sorry for being unclear. I meant "marginalizing" instead of "ignoring". It might make click what exactly means to marginalize / what is being marginalized.

ricardoV94: Fine by me then :)

bsmith89: <p>Oh! Thanks for the clarification.</p>

bsmith89: Cool, I think I addressed most of your comment on this text.

> It's not strickly one additional parameter right? It uses a different (random) parameter.

Was there something else you're getting at here? Maybe related to @AlexAndorra's suggestion that I better explain my `alpha = frac * conc` parameterization?

bsmith89:
I went through and made changes as suggested by you three, @AlexAndorra, @ricardoV94, and @MarcoGorelli . I then went ahead and "resolved" most of the inline conversations where I thought I had sufficiently addressed them.

The only two things that I've left open are
- the discussion about using the `arviz.plot_ppc` function instead of my bespoke code
- a question about better explaining my ad-hoc re-parameterization (for interpretability) of the Dirichlet and the DM to use a vector on the simplex (category probability) and a scalar (concentration), instead of the canonical k-dimensional alpha parameter.

@AlexAndorra, yours is the only unresolved code review. Once you're satisfied, I think this can be merged.

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
ricardoV94: <p>I still think the one extra parameter is misleading. While it's true that the Dirichlet has k free parameters and the DM has k-1 (because we normalize them), this is not where it's flexibility is coming from. You can model the Multinomial with a single Dirichlet prior on the p and it still won't be as flexible as the DM. It's the fact that each batch of observations is allowed a different draw from this Dirichlet that makes it more flexible.</p>

ricardoV94: <p>Also the concentration is not a parameter of the general DM (it is of your particular data generating model below). I wouldn't mention it yet. Specially because it sounds like this is the"extra" parameter that distinguishes the DM from the Multinomial.</p>

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
```diff
@@ -0,0 +1,1815 @@
...
```
ricardoV94, line 1: <p>Simplify to "Thankfully, we only get one warning related to inefficient sampling"?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

ricardoV94, line 1: <p>raw&nbsp;multinomial -&gt; simple multinomial</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

ricardoV94:
@bsmith89 I think the last changes are really great. I left two minor sentence suggestions. 

More importantly I replied to our thread on the "extra" parameter of the DM. Let me know if I explained myself well.

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
AlexAndorra: <p>Yeah, I agree with Ricardo here: first introducing the DM as a generalization of Beta-Binomial (as you already started doing, this is just a small tweak, and as Ricardo does here: "It's the fact that each batch of observations is allowed a different draw from this Dirichlet that makes the DM more flexible than the Multinomial").</p><p>And then, later in the NB, better explaining your <code>alpha = frac * conc</code> parameterization -- when you actually use it I think is a good place</p>

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
AlexAndorra: <p>Again, it's because Metropolis isn't as good as NUTS to spot and warn about problematic geometries in the posterior. I'd reformulate as: "we only get one warning now, so it <em>seems</em> that our sampling was more efficient".</p><p>Also, I think you should show how this model runs with NUTS, so that people can see what inefficient sampling looks like</p>

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
AlexAndorra: <p>Super nice @bsmith89! Do you think you could also add the species in the legend (or elsewhere) for each plot? That way we know which color is what ;)</p><p> </p>

AlexAndorra: <blockquote>You're suggesting that I shouldn't use the latex π when the code has the word <code>pi</code>? </blockquote><p>Just saw that but yeah I agree: this is confusing for the reader, as I noted above</p>

ricardoV94: <p>Related to the <code>pi</code>  I was not having an issue with it, but I do agree about using the same. You can also use the π symbol in Python as a varible if you want to keep it the same in text and code.</p>

AlexAndorra (changes requested):
This is shaping up to be a really good example NB, I really love what you did here @bsmith89 🤩 
There is some context about the usefulness of DM at the beginning, and mention of its shortcomings at the end (and you don't just say "hey you can use something else than DM when that happens", you actually point to alternatives 👏 ).

I left a few comments and suggestions below. This should polish the NB even more, and once that is done I think we can merge 🥳

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
AlexAndorra, line 1: <p>Typo: "This example notebook demonstrates the use <strong>of</strong> a Dirichlet mixture of multinomials"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

AlexAndorra, line 1: <p>"Here we will pretend that we have observed counts of five different species of tree<strong>s</strong> in <s>a</s> <strong>several</strong> forest<strong>s</strong>." (if I understood correctly)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

AlexAndorra, line 1: <p>Just adding some comments to help readers:</p><pre class="ql-syntax" spellcheck="false">n_samples = 10 # nbr of forests observed
total_count = 50 # nbr of trees observed in each forest
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

AlexAndorra, line 1: <p>Add just a few words about how <code>\alpha</code> can be parametrized: all ones = uniform, and then the higher the number, the more weight on the corresponding category</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

AlexAndorra, line 1: <ul><li>By n, you mean n_samples here, right? If yes, would be better to use this denomination, to match the one in the code</li><li>"While either of these would probably overcome the problems with the default sampler": why is so?</li><li>"here we'll instead switch to a standard, Metropolis step method, increasing our number of draws to overcome the inefficiency relative to NUTS": this formulation is problematic to me, as the problem doesn't come from NUTS, it comes from the <em>model</em>. Switching to Metropolis doesn't make the model more efficient; it just makes the warnings go away, both because Metropolis just doesn't have them, and because the algorithm <em>itselft</em> is less robust. The current framing implies that switching to Metropolis makes the convergence problems <em>go away</em> when it just <em>hides</em> them. As it's a mistake that's often made by people coming to Bayes, I think we should change the formulation above. What do you think?</li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

AlexAndorra, line 1: <p>If you're on the latest ArviZ (0.11.0), we switched to <code>compact=True</code> by default ;)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

AlexAndorra, line 1: <p>Nice touch adding the mean ESS per second! Do you think it'd be useful to add that to ArviZ natively?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

AlexAndorra, line 1: <p>You can also use <code>az.plot_forest</code> to do that I think: https://arviz-devs.github.io/arviz/api/generated/arviz.plot_forest.html</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

AlexAndorra, line 1: <p>"where we have one anomalous observation of more than 20": I'd say it's more the model that has a problem accomodating, or expecting, this wide variation, than the observation being an anomaly -- if that's the true data generating process, that's not an anomaly</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

AlexAndorra, line 1: <ul><li>Nitpick: I'd call the concentration parameter... <code>concentration</code> , instead of <code>alpha</code>, which is more obscure</li><li>"In this iteration of our model we'll explicitly include the latent multinomial probability...": this is confusing: there is no <code>\pi</code> parameter in the model below. Didn't you rather mean:</li></ul><blockquote>In this iteration of our model we'll explicitly include the latent multinomial probability <code>p_i</code> , which models the probabilities of each species in each forest, and that we would never observe in the real world.</blockquote><p>?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

AlexAndorra, line 1: <p>Same here: there is no <code>\pi</code> in the model</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

AlexAndorra, line 1: <p>Same here: I think a forest plot would convey the point more clearly</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

AlexAndorra, line 1: <p>"we've collapsed the <strong>latent multidimensional</strong> Dirichlet and the multinomial nodes". As there are two Dirichlets in the model, that makes the distinction clearer</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

AlexAndorra, line 1: <p>Great plot, and amazing convergence and sampling with the DM model!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/18/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='41'/>

ricardoV94: <p>Actually the better would be to use the same parameter names you mentioned in the beginning when you generated our data, although then you have the problem of shadowing the original variables (I always struggle with this!). I usually end up doing <code>true_frac = ...</code>  when generating data, and in the models call it <code>frac</code>  <span class="ql-cursor">﻿</span></p>

bsmith89: <blockquote>Do you think it'd be useful to add that to ArviZ natively?</blockquote><p>...Meh. It's unfortunately very dependent on other stuff, like burn-in, what architecture you're on, chains vs. cores, etc. In the end, just comparing sampling time is probably the better option.</p><p> </p><p>Might be worth thinking about some more comparable measure. (Floating-point operations per ES? Is that possible?)</p>

AlexAndorra: Ah yeah, good point 🤔 However, if you think of helpful measures that would help diagnose models, feel free to open issues (or even PRs) on the [ArviZ](https://github.com/arviz-devs/arviz) side!

ricardoV94: Some smooth recruiting going on here ^^. I approve

AlexAndorra: Don't blow my cover @ricardoV94! 🤫

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
bsmith89: <p>Just pushed new variable names across the board.</p><p> </p><p>Now I'm using the same variable names in the data simulation as in the model (with <code>true_</code>  prepended), and I've dropped single character symbols throughout (except for indexing and shapes). In text I use latex for all of the variable names.</p><p> </p><p>This means that I'm basically ignoring the distinction between true parameters and model parameters, which is probably the right move for this notebook.</p>

bsmith89: <p>I've updated this text (and the explanation of the simulation below). Does the current iteration address your concerns about the reparameterization and the explanation for the DM's flexibility?</p>

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
bsmith89: <p>:P</p>

bsmith89: <p>I've renamed simulation and model variables throughout. See my comment above.</p>

bsmith89: <p>Addressed. See comment above.</p>

bsmith89: <blockquote> Great plot</blockquote><p>Thanks! Just fancy-fied it even more with KDEs of the posterior samples and legends.</p><p> </p><p>Thoughts? Too busy? Should the discussion/visualization of biased parameter estimation be split out entirely to a forest plot?</p>

bsmith89: <blockquote>Also, I think you should show how this model runs with NUTS, so that people can see what inefficient sampling looks like</blockquote><p> </p><p>Unfortunately, NUTS fails entirely with:</p><p> </p><pre class="ql-syntax" spellcheck="false">---------------------------------------------------------------------------
RemoteTraceback                           Traceback (most recent call last)
RemoteTraceback: 
"""
Traceback (most recent call last):
  File "~/Projects/pymc3/pymc3/parallel_sampling.py", line 191, in _start_loop
    point, stats = self._compute_point()
  File "~/Projects/pymc3/pymc3/parallel_sampling.py", line 216, in _compute_point
    point, stats = self._step_method.step(self._point)
  File "~/Projects/pymc3/pymc3/step_methods/arraystep.py", line 276, in step
    apoint, stats = self.astep(array)
  File "~/Projects/pymc3/pymc3/step_methods/hmc/base_hmc.py", line 159, in astep
    raise SamplingError("Bad initial energy")
pymc3.exceptions.SamplingError: Bad initial energy
"""

The above exception was the direct cause of the following exception:

SamplingError                             Traceback (most recent call last)
SamplingError: Bad initial energy

The above exception was the direct cause of the following exception:

ParallelSamplingError                     Traceback (most recent call last)
&lt;ipython-input-22-16dea9c651bc&gt; in &lt;module&gt;
      1 with model_multinomial:
----&gt; 2     trace_multinomial = pm.sample(
      3         draws=int(5e3), chains=4, return_inferencedata=True
      4     )

~/Projects/pymc3/pymc3/sampling.py in sample(draws, step, init, n_init, start, trace, chain_idx, chains, cores, tune, progressbar, model, random_seed, discard_tuned_samples, compute_convergence_checks, callback, jitter_max_retries, return_inferencedata, idata_kwargs, mp_ctx, pickle_backend, **kwargs)
    555         _print_step_hierarchy(step)
    556         try:
--&gt; 557             trace = _mp_sample(**sample_args, **parallel_args)
    558         except pickle.PickleError:
    559             _log.warning("Could not pickle model, sampling singlethreaded.")

~/Projects/pymc3/pymc3/sampling.py in _mp_sample(draws, tune, step, chains, cores, chain, random_seed, start, progressbar, trace, model, callback, discard_tuned_samples, mp_ctx, pickle_backend, **kwargs)
   1473         try:
   1474             with sampler:
-&gt; 1475                 for draw in sampler:
   1476                     trace = traces[draw.chain - chain]
   1477                     if trace.supports_sampler_stats and draw.stats is not None:

~/Projects/pymc3/pymc3/parallel_sampling.py in __iter__(self)
    477 
    478         while self._active:
--&gt; 479             draw = ProcessAdapter.recv_draw(self._active)
    480             proc, is_last, draw, tuning, stats, warns = draw
    481             self._total_draws += 1

~/Projects/pymc3/pymc3/parallel_sampling.py in recv_draw(processes, timeout)
    357             else:
    358                 error = RuntimeError("Chain %s failed." % proc.chain)
--&gt; 359             raise error from old_error
    360         elif msg[0] == "writing_done":
    361             proc._readable = True

ParallelSamplingError: Bad initial energy
</pre><p> </p><p>I messed around with <code>target_accept</code>  and <code>testval</code>,  but I couldn't get all 4-chains to finish without numerical issues.</p>

bsmith89: <p>I haven't drilled down into the NUTS sampling errors, but I'm guessing that we're getting logp's or dlogp's of <code>-inf</code> due to the large deviation from the multinomial predictive density (e.g. maybe <code>observed_counts[0, 4] == 4</code> is just too unbelievable?).</p><p> </p><p>Since Metropolis either (A) rejects jumps into regions of the parameter space where logp is <code>-inf</code>, or (B) doesn't know about the numerically unstable gradient, it doesn't fail the same way NUTS does.</p><p> </p>

bsmith89: <p>I haven't drilled down into the NUTS sampling errors, but I'm guessing that we're getting logp's or dlogp's of&nbsp;<code>-inf</code>&nbsp;due to the large deviation from the multinomial predictive density (e.g.&nbsp;<code>observed_counts[3, 1] == 1</code>&nbsp;is pretty improbable given that the expectation is 15).</p><p> </p><blockquote>Switching to Metropolis doesn't make the model more efficient; it just makes the warnings go away, both because Metropolis just doesn't have them, and because the algorithm&nbsp;<em>itselft</em>&nbsp;is less robust. The current framing implies that switching to Metropolis makes the convergence problems&nbsp;<em>go away</em>&nbsp;when it just&nbsp;<em>hides</em>&nbsp;them.</blockquote><p> </p><p>Since Metropolis either (A) rejects jumps into regions of the parameter space where logp is&nbsp;<code>-inf</code>, or (B) doesn't even know about the numerically unstable gradient, it doesn't fail the way NUTS does.</p><p> </p><blockquote>"While either of these would probably overcome the problems with the default sampler"</blockquote><p> </p><p>Haven't actually tried it, but I'm conjecturing here that the <s>Bernoulli</s> categorical likelihood elements (note: Bernoulli was a mistake) won't be as "surprised" (so to say) with hits to the low frequency species since any one observation is fairly believable. Equivalently with the collapsed likelihood, since there's nothing surprising about any combination of species counts. (...Does this explanation make any sense?)</p><p> </p><p>(Addendum: I went ahead and tried out the collapsed model, which works the way I expected.)</p><p> </p><blockquote>this formulation is problematic to me [...] As it's a mistake that's often made by people coming to Bayes, I think we should change the formulation above.</blockquote><p> </p><p>Do you have a suggestion for better wording?</p><p> </p><blockquote>What do you think?</blockquote><p> </p><p>Mmm...I'm pretty okay with mostly sweeping these issues under the rug in this example. I think the details you're considering here should be in an e.g. "Diagnosing Biased Inference" or "Reparameterizing Problematic Models" notebooks. Here that problem serves as good motivation for using a better model, but I don't think we need to deal with it much more than that.</p>

bsmith89: <p>Yep, added.</p>

bsmith89: <p>Added.</p>

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
bsmith89: <p>Labels added as legend titles.</p>

ricardoV94: <p>I think the new text is much better. Now I would only cut back a bit on redundancy: </p><p> </p><blockquote>Unlike the <a href="https://en.wikipedia.org/wiki/Multinomial_distribution" target="_blank">Multinomial distribution</a>, in which the variance of counts is entirely  determined by the sample size and the probability of each category,  the DM is more flexible than the multinomial, allowing it to  model counts that are <em>more variable</em> than the Multinomial  would predict.  The Dirichlet-multinomial can be understood as draws from the multinomial  but where each sample has a slightly different probability vector,  which is itself drawn from a common <a href="https://en.wikipedia.org/wiki/Dirichlet_distribution" target="_blank">Dirichlet distribution</a>. </blockquote><p>I think you don't need to mention the sample size parameter, as this one works exactly the same in the M and DM. I think something shorter like this might do the job better:</p><p> </p><blockquote>The Dirichlet-multinomial can be understood as draws from a Multinomial distribution where each sample has a slightly different probability vector, which is itself drawn from a common <a href="https://en.wikipedia.org/wiki/Dirichlet_distribution" target="_blank">Dirichlet distribution</a>. This contrasts with the Multinomial distribution, which assumes that all observations arise from a single fixed probability vector. This enables the Dirichlet-multinomial to accommodate more variation in count data than the Multinomial. </blockquote>

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
ricardoV94: <p>I think, as you said, it's now better to remove the posterior and true. The forest plots illustrate it perfectly, and this one looks a bit too busy with everything in it. The new legend works nicely btw.</p>

ricardoV94: <p>I agree that it might be a bit out of the scope of this Notebook to explore why NUTS fails. Maybe just warn more explicitly that this is not a real solution to the problem, but just a shortcut so you have something to compare with the proper DM model?</p>

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
AlexAndorra: <p>Yep, all good now, thanks! I also like Ricardo's proposal -- it's up to you</p>

examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
AlexAndorra: <blockquote>I'm guessing that we're getting logp's or dlogp's of&nbsp;<code>-inf</code>&nbsp;due to the large deviation from the multinomial predictive density (e.g.&nbsp;<code>observed_counts[3, 1] == 1</code>&nbsp;is pretty improbable given that the expectation is 15).</blockquote><p>Yeah, that'd be my guess too. We could probably help NUTS run by selecting better priors than the current uniform ones, but, by design, the issue is deeper -- the model is just not a good enough approximation of the generative process.</p><p> </p><blockquote>Does this explanation make any sense?</blockquote><p>Yeah thanks, I think I understand what you mean. You can see if you can briefly add that in the text above, but it's not a blocker.</p><p> </p><blockquote>'m pretty okay with mostly sweeping these issues under the rug in this example. [...] Here that problem serves as good motivation for using a better model, but I don't think we need to deal with it much more than that.</blockquote><p>I agree with that. What I'm saying is we should say that doing so is not <em>fixing</em> issues but <em>sweeping</em> them under the rug, otherwise readers could think that just switching to Metropolis when NUTS breaks is a valid workflow, whereas the latter happening should be a red alert -- actually, I think there'd be value in mentioning or even showing that NUTS doesn't even want to sample such a model --&gt; folk theorem must pop in our head.</p><p> </p><blockquote>Do you have a suggestion for better wording?</blockquote><p>The way you reworded it is already quite good! I'd just add a bit to it. Something like:</p><p>"... here we'll instead switch to the Metropolis step method, which ignores some of the geometric pathologies of our naïve model. This is important: switching to Metropolis does not not <em>fix </em>our model's issues, rather it <em>sweeps them under the rug</em>. In fact, if you try running this model with PyMC's default sampler, NUTS will break loudly from the beginning. When that happens, this should be a red alert that there is something wrong in our model.</p><p>You'll also notice below that we have to increase considerably the number of draws we take from the posterior, because Metropolis is much less efficient than NUTS."</p><p> </p><p>Ricardo's last comment is also spot-on:</p><blockquote>And later don't emphasize so much that the diagnostics look nice, but caution again that while they may look nicer we are probably glossing over very real problems in the posterior approximation that the NUTS made clear.</blockquote><p> </p><p> </p>

AlexAndorra: <p>I wrote that above: I think the posterior can be removed but the true values don't hurt as those are just lines</p>

AlexAndorra (changes requested):
Thanks a lot for the revisions @bsmith89 ! I answered to a few of your comments and left another few -- we're getting there 💪

bsmith89:
I'm hoping that's everything.

🤞 !

ricardoV94:
I

> I'm hoping that's everything.
> 
> 🤞 !

Looks really great! Nice work. I have no more suggestions :). It's up to our editor @AlexAndorra now

ricardoV94 (approved):

AlexAndorra (approved):
Yes, all good now 🥳 Thanks a lot for sticking to it @bsmith89, this is a great example!

Issue #2, opened 2020-12-15
Use sigma instead of sd


MarcoGorelli:
e.g. `Normal('x', mu=mu, sigma=sigma)` rather than `Normal('x', mu=mu, sd=sigma)`

`sd` will (silently) continue working, see https://github.com/pymc-devs/pymc3/pull/4344

CloudChaoszero:
Hey @MarcoGorelli, I am working on some big changes related to switching from pymc3 plots to arviz plots. 
* Being said, I can take this on as well, and work on both of them.

twiecki:
That's great @CloudChaoszero - much appreciated!

CloudChaoszero:
I updated the PR to include replacing `sd` with `sigma` parameter. 
* If any lingering ones exist, we can update it over time, I imagine.

CloudChaoszero:
I think this Issue can be closed now. 🤞

PR #16, opened 2021-01-04
Replacing PyMC3 plots w/ Arviz plots & sigma Param change

CloudChaoszero:
# Description

Replace PyMC3 dependent plots with arviz plots in case studies & examples.
* The issue is a 3-part step for the following:
  * Initiative via Issue https://github.com/pymc-devs/pymc3/issues/4371 (Leave open)

Replace parameter `sd` with `sigma` (e.g. some examples have `pm.Normal(...sd=...)`
* Closes issue #2

PR #24, opened 2021-01-19
Replacing PyMC3 plots w/ Arviz plots & sigma Param change [Part 4]

CloudChaoszero:
# Description
The following is a large PR breakdown of PR #16.

Replace PyMC3 dependent plots with arviz plots in case studies & examples.
* The issue is a 3-part step for the following:
  * Initiative via Issue https://github.com/pymc-devs/pymc3/issues/4371 (Leave open)

Replace parameter `sd` with `sigma` (e.g. some examples have `pm.Normal(...sd=...)`
* Closes issue #2


OriolAbril:
> In most other notebooks we have the watermark at the end?

Yes, all notebooks are supposed to have the watermark at the end.

> The interactive plots are always a bit dangerous, because they often get lost when saving the notebook

Agreed, however, I don't think there are interactive plots in the notebook

> `KeyError: 'var names: "['sd_log__'] are not present" in dataset'`

This is because ArviZ does not store the transformed variables in the `posterior` group. You can either use the transform argument in the plotting functions (I think they all have it but not sure). Or manually add the transformed variable to the dataset. See one example here: https://github.com/arviz-devs/arviz/issues/641

OriolAbril:
Maybe we can create an issue as a remainder to fix this but for now go ahead and merge the PR?

CloudChaoszero:
> Maybe we can create an issue as a remainder to fix this but for now go ahead and merge the PR?

Sounds good @OriolAbril ! I made [this issue](https://github.com/pymc-devs/pymc-examples/issues/43)

OriolAbril:
I saw the issue, thanks! I don't see the comments of this PR on the issue yet, especially the ones about the glm comparison one, can you add them?

CloudChaoszero:
> I saw the issue, thanks! I don't see the comments of this PR on the issue yet, especially the ones about the glm comparison one, can you add them?

Sounds good. @OriolAbril ! I added it at the last section, of the original description, and linked one of this PR's comments

PR #36, opened 2021-02-14
Delete all notebooks

Pyrsos:
Create example repo branch, excluding all notebook files to allow for faster pymc3 build.


MarcoGorelli:
Thanks @Pyrsos - I've made a new `pymc-devs:fast-docs-build` branch for this, and will merge into that one (rather than `main`, from which the notebooks definitely should not be removed :smile: )

Issue #33, opened 2021-02-04
GLM in PyMC3: Out-Of-Sample Predictions Example


juanitorduz:
Hi 👋 ! I am very interested in the `glm` module of PyMC3, specially using it for prediction problems. There is an active thread about this topic: [“Out of sample” predictions with the GLM sub-module](https://discourse.pymc.io/t/out-of-sample-predictions-with-the-glm-sub-module/773). I wrote a small post summarising the potential solutions : https://juanitorduz.github.io/glm_pymc3/  Is this something you would be interested in having on the [examples](https://docs.pymc.io/nb_examples/index.html)? If so, is it ok as it is now or do you want me to reduce it a just focus on the out-of-sample section. I am happy to contribute to the documentation 🚀 .

ricardoV94:
@juanitorduz, sorry for the delay! From checking that discourse thread, it seems there is a strong need for a concise example on how to do out of sample predictions for GLM models. 

I think your blog does a good job so it would probably be a good addition to the examples gallery. 

I would think that we might already have some examples buried in other longer glm notebooks, so if you could link to those at the end of your notebook that would be great (even if they don't use Patsy). If we really have no other examples, that makes an even stronger care for your addition.

Pinging @AlexAndorra for input.

juanitorduz:
I created an initial PR https://github.com/pymc-devs/pymc-examples/pull/37

PR #37, opened 2021-02-14
Add GLM Out of Sample Predictions Notebook 

juanitorduz:
I would like to add a small example to the docs. See https://github.com/pymc-devs/pymc-examples/issues/33 and the original post https://juanitorduz.github.io/glm_pymc3/. I am open for suggestions and changes to make this notebook useful for the community.


juanitorduz:
> View / edit / reply to [this conversation](https://app.reviewnb.com/pymc-devs/pymc-examples/pull/37/discussion/) on ReviewNB
> 
> _MarcoGorelli commented on 2021-02-23T14:55:33Z_
> ----------------------------------------------------------------
> 
> > This could be the case when one-hot-encoding categorical variables.
> 
> This isn't something I've come across, and I'm not sure I see how that would happen as the one-hot encoding using training set information only - do you have a reference?

This is related to (this post)[https://dzone.com/articles/pandasscikit-learn-get-dummies-testtrain-sets], where once the model is deployed it might see new categories. One could try to use `handle_unknown = 'ignore'` if this is expected (and wanted), but then in the cross-validation step if we use `get_dummies` we are already showing the model what to expect. Maybe I can just delete this line as it is not very relevant for the post itself :)

juanitorduz:
> View / edit / reply to [this conversation](https://app.reviewnb.com/pymc-devs/pymc-examples/pull/37/discussion/) on ReviewNB
> 
> _MarcoGorelli commented on 2021-02-23T14:55:34Z_
> ----------------------------------------------------------------
> 
> are you sure that
> 
> fpr, tpr, thresholds = roc_curve(
>     y_true=y_test, y_score=y_test_pred, pos_label=1, drop_intermediate=False
> )
> is correct? Looking at the sklearn docs , it seems that the second argument (y_score) should be the raw predictions p_test_pred rather than the predicted labels y_test_pred . Indeed, if you look at the examples in the sklearn docs, they pass `clf.predict_proba(X)[:, 1]`

You are totally right 🙈 ! Thank you so much! I corrected the typos and this error on [this commit](https://github.com/pymc-devs/pymc-examples/pull/37/commits/11be4ae6c3484cbf9b700c744204087496fe8a9e).

MarcoGorelli:
Cool, thanks @juanitorduz - looks good to me, I would just advise to add a prior predictive check (even though it's not the main focus of your notebook, it's probably something to encourage where possible)

juanitorduz:
@MarcoGorelli I added a subsection on prior predictive checks. Let me know what you think about it 🤓 .

juanitorduz:
Thanks @OriolAbril ! I fixed it :)

examples/table_of_contents_examples.js
```diff
@@ -17,6 +17,7 @@ Gallery.contents = {
     "case_studies/hierarchical_partial_pooling": "(Generalized) Linear and Hierarchical Linear Models",
     "generalized_linear_models/GLM-model-selection": "(Generalized) Linear and Hierarchical Linear Models",
     "generalized_linear_models/GLM-negative-binomial-regression": "(Generalized) Linear and Hierarchical Linear Models",
+    "generalized_linear_models/GLM-out-of-sample-predictions": "Generalized) Linear and Hierarchical Linear Models",
```
twiecki, line 20:
```suggestion
    "generalized_linear_models/GLM-out-of-sample-predictions": "(Generalized) Linear and Hierarchical Linear Models",
```

juanitorduz:
Thank you @ricardoV94 I fixed your suggestions! (I always have issues naming `patsy` 😆 )!

juanitorduz:
Thank you all for  your suggestions and catching my typos 🙈  (it is hard to catch them with a 1-year old learning to walk around you 😄 )!

twiecki:
@MarcoGorelli Yeah, when I read that I thought "this is a great suggestion" and I love prior checks but now seeing it implemented I don't think it really benefits much here and I can't think of a way to improve it, so I would argue to remove it again. What do you think?

ricardoV94 (changes requested):
I left a comment about the true data model specification. It's not a blocker, but I would at least mention it in the notebook.

ricardoV94 (approved):
LGTM

twiecki:
Thanks @juanitorduz!

PR #39, opened 2021-02-23
Fixup putting workflow

MarcoGorelli:
fixing it up to use inference data and ArviZ

Addresses issue #50.


twiecki (approved):

twiecki:
Is it worth to resolve these conflicts? The changes here make the NB better.

MarcoGorelli:
Thanks! I'll fix it up, hope to get back to this soon-ish

OriolAbril:
I would rather have this changes than the current ones, if it's easier I think you can just ignore the current changes and use your version.

MarcoGorelli:
> I would suggest using xarray for computation here

Thanks for your suggestion! I haven't yet really made sense of xarray and its magic, so for now I've just resolved the conflicts (i.e. `git merge upstream/main --strategy-option ours` - I'd repeated many of the changes from the sd->sigma PR here anyway)

MarcoGorelli:
Something I don't understand about this notebook is that there's a cell which says

```
The lesson from this is that

$$
\mathbb{E}[f(\theta)] \ne f(\mathbb{E}[\theta]).
$$

this appeared here in using 

```python
# Right!
scipy.special.expit(logit_trace_combined["a"] * 50 + logit_trace_combined["b"]).mean()
```
rather than

```python
# Wrong!
scipy.special.expit(logit_trace_combined["a"].mean() * 50 + logit_trace_combined["b"].mean())
```

to calculate our expectation at 50 feet.
```

But then, it proceeds to do
```python
ax.plot(t, forward_angle_model(angle_trace["variance_of_shot"].mean(), t))
```
Based on the above, I think I would've expected something like
```python
ax.plot(t, forward_angle_model(angle_trace["variance_of_shot"], t).mean())
```

MarcoGorelli:
not sure why the diff isn't being picked up nicely by review-nb, will take another look later

OriolAbril:
> not sure why the diff isn't being picked up nicely by review-nb, will take another look later

Don't worry about it, it happens quite often (at least it has happened quite often in PRs I review) reviewnb just misses the point completely from time to time and thinks the whole notebook was updated.

OriolAbril:
> However, based on the above, I think I would've expected something like
>
>     ax.plot(t, forward_angle_model(angle_trace["variance_of_shot"], t).mean())

I don't know what `forward_angle_model` does, but if this work we should use this and always leave the averaging for last.

OriolAbril (approved):

MarcoGorelli:
Thanks @OriolAbril for your really helpful review!

Issue #14, opened 2020-12-27
Change examples to use coords where appropriate


twiecki:
`coords` is a very useful feature that not enough people know about. Our docs should establish best practices so using it in our NBs is an important step.

AlexAndorra:
+1 💯 
This is also a good way to introduce more people to the [wonderful `InferenceData` backend](https://arviz-devs.github.io/arviz/getting_started/XarrayforArviZ.html), which will be returned by default by `pm.sample` in the next version of PyMC3 👌

twiecki:
@CloudChaoszero This is getting a lot but I wonder if at least with the changing of the examples in https://github.com/pymc-devs/pymc-examples/pull/16 we should at least have them use `InferenceData` (using coords would be the cherry).

CloudChaoszero:
> @CloudChaoszero This is getting a lot but I wonder if at least with the changing of the examples in #16 we should at least have them use `InferenceData` (using coords would be the cherry).

Sounds great @twiecki! I saw this earlier, but was unsure what `coords` was (i skimmed Alex's link lol). 

Is that another term (or method of said class) `InferenceData`?

twiecki:
@CloudChaoszero No, it's a way to specify dimensions in a model. For example, for Radon, instead of just specifying `shape=len(counties)`, you could specify `dims='counties'` and then also have the names of the counties show up in the `InferenceData` object, rather than just the numbers, which also makes plotting easier.

I do have to warn you though, this would require you to level up on xarray and arviz and would be quite an important but also time-intensive refactor. You can just take it one NB at a time and stop when you're bored, every little bit helps, but just wanted to mention that. Alternatively, just changing the plotting and stats functions to arviz alone would already be a great contribution, and probably more important in terms of time-line.

AlexAndorra:
Together with `dims`, this is what allows more intuitive multi-dimensional data indexing with `InferenceData` (instead of plain numpy-array indexing, which is painful for humans for multi-dim data, as I'm sure you already noticed 😅 ).

You can see that in the link I posted above: the posterior data has 3 `dims`, `chain`, `draw` and `school`, which respectively contain 4, 500 and 8 `coords`. For `chain`, those coordinates are 0, 1, 2 and 3 (because no need to name the chains), but for `school` they have informative labels: 'Choate', 'Deerfield', 'Phillips Andover', 'Phillips Exeter', ... 
[...]

PR #45, opened 2021-03-21
Update GLM hierarchical binomial model to best practices

CloudChaoszero:
This PR is the start of https://github.com/pymc-devs/pymc-examples/issues/14

Addresses #61 and updates the notebook to "Best Practices"

## Summary

* Have trace objects saved as InferenceData objects via `return_inferencedata=True`.
* If an opportunity for using `coords` exists, then add param `coords`

I'm open to interpretation on if this is what should be the goal for linked issue  https://github.com/pymc-devs/pymc-examples/issues/14.


CloudChaoszero:
Performing required updates to 1-2 files now. 
* Gauging if these are the ideal changes needed.

OriolAbril:
Also, minor note, I think you'll run into merge issues with #24 here.

OriolAbril:
There are merge conflicts now :sweat_smile:

CloudChaoszero:
@OriolAbril Yikes ha-- updated now!

CloudChaoszero:
## Sigma = 5




## Sigma = 10

OriolAbril:
Thanks for the plots with multiple sigmas, it looks like from 5 to 10 it still changes significantly, can you take a look at sigma=20 and sigma=30? no need to post them, only check the intuition that the posterior progressively stops being so influenced by the prior. We are already retrieving the expected values with 10 so everything looks good. Thanks again for the PR :smile:

CloudChaoszero:
> Thanks for the plots with multiple sigmas, it looks like from 5 to 10 it still changes significantly, can you take a look at sigma=20 and sigma=30? no need to post them, only check the intuition that the posterior progressively stops being so influenced by the prior. We are already retrieving the expected values with 10 so everything looks good. 

@OriolAbril  Yeah, so there isn't a significant change in the posteriors after updating to sigma =20, sigma=30, seen below.

And no problem! 😄  This PR gives me some guidance on fully implementing `return_inferencedata` and `coords` changes for this examples repo.
______

## Sigma = 20

Screen Shot 2021-04-04 at 1 45 14 AM


## Sigma = 30

Screen Shot 2021-04-04 at 1 48 39 AM

OriolAbril (approved):

Issue #49, opened 2021-03-24
rugby analytics


OriolAbril:
file: https://github.com/pymc-devs/pymc-examples/blob/main/examples/case_studies/rugby_analytics.ipynb

## Required changes
The inital exploratory analysis looks ok, seaborn is an ok dependency for this, the model needs to be updated, no flat priors for example and we should also use coords+dims and ArviZ for posterior analysis and exploration.

*Notes:* updates on the model and priors can be taken from https://github.com/arviz-devs/arviz_example_data/blob/main/rugby.ipynb.

almostmeenal:
@OriolAbril can I take this

canyon289:
Assigned to this as well. I would suggest closing the two you're assigned to before taking on any more PRs

almostmeenal:
ran into some error- https://github.com/arviz-devs/arviz/pull/1117#issuecomment-808915434 , also random but would it be a good idea to possibly create a similar example notebook with something like ucl or something, with recent data and current best practices, I saw that this notebook didnt even import arviz. can something be done to ensure examples get updated accordingly each time a new PR is merged which revamps/removes/modifies/adds a new feature or utility to ensure they stay updated. I mean, as contributors yeah we can look up errors and everything but for users who want to learn from examples repo, running into errors there is kind of hard to fix.

twiecki:
@almostmeenal Yes, I think updating the NBs to use best current practices is important, like arviz and `coords`.

almostmeenal:
@OriolAbril hi so I'm thinking a lot of things and I'm slightly lost wrt what exactly am i supposed to do, so I'll start a pull request while I'm working on it so you can maybe help me out a bit, from what I think, I guess I'm supposed to change the part after building the model and use arviz inference data format for data and arviz plots instead of the ones used. and for building the model take the part from rugby.ipynb in arviz_example_data

OriolAbril:
Sending a PR sounds great! I have been adding more and more guidance and content but I can't go as fast as I'd like. For now you'll see I have already added a [contributing guide](https://github.com/pymc-devs/pymc-examples/blob/main/CONTRIBUTING.md) that is specific to pymc-examples and when you submit the PR, you'll see I have also added a pull request template. 

> I guess I'm supposed to change the part after building the model and use arviz inference data format for data and arviz plots instead of the ones used

yes, these changes would "move" the notebook from "To Do" to "ArviZ" column.

[...]

PR #62, opened 2021-03-30
modified rugby analytics case study with increased arviz usage

almostmeenal:
I tried to convert some of the plots to arviz wherever possible, removed a few warnings because of deprecated plot names, and added inference data as arviz one
still working on: adding priors etc

Addresses issue #49  and aims to advance it to "Best Practices" column


almostmeenal:
@OriolAbril  I had a few doubts

1. in the second last cell i tried:
  az.plot_pair(data,
  var_names=['atts'],
  coords = {"team": teams.team.values},
  kind='scatter',
  divergences=True,
  textsize=25),
  figsize=(10, 10)
  instead of sns.pairplot
  does that make sense?

2. In the plot " HDI of Attack Strength, by Team", can we again use arviz forestplot?
3. Now are there any other changes I should look into other than adding prioirs?

OriolAbril (commented):
I added several comments with review nb, that are also added to github, but don't read them on github as there its not possible to see to which cells and changes they are referring to.

almostmeenal:
@OriolAbril i couldnt make a commit because of black failing as a precommit hook, i'll look into it and resolve it, then you can check the cells, just give me some time

OriolAbril:
You should try recomitting again right after the first time when it fails. `pre-commit` should _edit_ the files in addition to making the commit error out, so the 2nd time you try requirements should be met, at least for black which is completely automated. 

You can also run `pre-commit` manually as explained in the [jupyter style guide](https://github.com/pymc-devs/pymc3/wiki/PyMC3-Jupyter-Notebook-Style-Guide)

almostmeenal:
@OriolAbril okay so now its done i think you can check now, ive commented the simtable part cell, i went through that notebook you suggested but i think its doing something different w the log likelihood

OriolAbril:
> yeah i thought that was supposed to happen, so when i tried the next time it showed the same error again, when i pushed it just showed everything up to date

Oh, I see, yes, I said commit above but you first have to make sure to _add_ the modified files before committing again, or use `git commit -a` (a for all) which is what I tend to do and forget to be explicit on the add part. In any case, take a look at `git status` right after running pre-commit to see if you have changes not staged to commit.

> I've used arviz inference data to extend the posterior predictive part as you mentioned, but I've stuck to dataframes for generating the sim table, I tried using xarray operations to calculate 3*home_won and do the grouping, it was getting complicated(maybe its just me). But i think its fairly reasonable to use pandas for generating that table, as all the data stays in arviz.inferencedata but its manipulation for viewing can be done using pandas.
>
> Also, i kept constant data and calculated it with pp, because I feel like it's a good practice in general, for simple manipulation of arviz data, having the constant part contains which is home_team etc which should be stored too, makes it all happen in arviz and reduces the need to get data from initial csv/df.
> 
> Let me know if i should change anything else

I have skimmed through the changes and they look amazing, thanks! There are some nits to polish, but nothing to big. I'll review later on reviewnb.

Regarding the sim table, if you don't mind I can give it a go, now that everything else uses xarray and ArviZ it will be much easier for me to jump in and focus only on this problematic computation. Using pandas is fine, but I think that here everything will be much simpler once converted to xarray because the data in inherently multidimensional and trying to get pandas to play nicely with this data takes more time than the job itself. If so, I would do this after giving you time to work on the minor changes still needed so there are no git conflicts and push directly to this PR.

almostmeenal:
that sounds great!!

almostmeenal:
I was just exploring the xarray thing and the rank function needs bottleneck as a requirement, so i guess it wouldn't be a good idea to have another pre requisite?

OriolAbril:
> I was just exploring the xarray thing and the rank function needs bottleneck as a requirement, so i guess it wouldn't be a good idea to have another pre requisite?

I did not know about bottleneck, it looks useful but I don't know how easy it would is to install, as it is an optional dependency of xarray, I don't think it's an issue to require having it for the notebook as long as we document that. If think it can give problems and would rather not have it, we can use `scipy.stats.rank` together with `xarray.apply_ufunc` and not require anything. I trust your judgement here.

> instead of the 3/4 sim table cells,I'm getting the same barplot as before, so that works fine, I think the sim_table isn't very useful anyway and maybe if we do away with that then this seems fine?
I tried using xarray.count() but that didnt work so i made a loop and counted list occurences.

The code above looks like a good start, and is much simpler than the current version. I think it is not doing the same thing though, it seems like it counts the total of goals instead of counting the number of wins. 

> if you think this makes sense, then I'll make a commit else you can take it up, whatever seems appropriate

Whatever you prefer and you think you'll enjoy more, playing around with xarray groupbys and ranks is not very related to timeseries which I think is the project you were interested in, and you have already done an amazing job so I don't mind taking over or continue helping.

almostmeenal:
> I did not know about bottleneck, it looks useful but I don't know how easy it would is to install, as it is an optional dependency of xarray, I don't think it's an issue to require having it for the notebook as long as we document that. If think it can give problems and would rather not have it, 

I think we can have it, the installation was as simple as: `pip install bottleneck`

> The code above looks like a good start, and is much simpler than the current version. I think it is not doing the same thing though, it seems like it counts the total of goals instead of counting the number of wins.

yeah, I noticed that. I picked and modified the groupby part from your log_lik notebook. I was still at crossroads, whether calculating wins was actually doing anything informative in terms of how pymc3/arviz or bayesian stats work. I'm not sure, but would it make sense to just make that bar plot of probability of maximum number of _goals_ instead of _points_ . 

> Whatever you prefer and you think you'll enjoy more, playing around with xarray groupbys and ranks is not very related to timeseries which I think is the project you were interested in, and you have already done an amazing job so I don't mind taking over or continue helping.

I understand it won't be relevant for the time series project, but I had fun learning about xarrays, it felt like a more complicated but functional version of pandas. I think it might take a few mistakes, but I'll get there so I'm interested in working on it, it'll be great if you'd help me with that!! otherwise ofcourse its your project and if at some point you feel like I'm struggling a lot and taking up too much time you can take it up then.

OriolAbril:
> yeah, I noticed that. I picked and modified the groupby part from your log_lik notebook. I was still at crossroads, whether calculating wins was actually doing anything informative in terms of how pymc3/arviz or bayesian stats work. I'm not sure, but would it make sense to just make that bar plot of probability of maximum number of goals instead of points .

calculating on the number of goals would answer a different question, for what we are interested in here, we need to use the number of points because it's the quantity that defines who wins the league and using goals can give a different answer. One example were using goals would be misleading is if one team lost all matches by a single goal except one where they won by 50 goals. They would probably win in terms of goals (losing by 1 leaves them relatively close to all the other teams, and 50 is a significant advantage), but they would only have 3 points! A team that instead won all games but only by a 1 goal difference would have 18 points even if they lost in terms of total goals.

This was done in the first cell with the dict comprehensions, but it can be done in a vectorized way with xarray. We can then groupby and sum on the boolean dataarray we have obtained from evaluating the condition instead. I also noticed that teams get 3 points like it happens in football, which is not what happens in rugby. I don't really know about rugby, and from some [wikipedia](https://en.wikipedia.org/wiki/2021_Six_Nations_Championship#Table) reading we don't have enough information to get the actual score, but we can definitely take ties into account for example or look at differences of less than 7 goals.

almostmeenal:
> This was done in the first cell with the dict comprehensions, but it can be done in a vectorized way with xarray.

I was trying to do that with xarray.where and that didnt work out so well, and I'm guessing using recurring for loops would again be a redundant idea

OriolAbril:
> I was trying to do that with xarray.where and that didnt work out so well

Given that we only want the boolean result, you should be able to evaluate the condition directly: `(pp["home_points"] > pp["away_points"]) * 3` should return an interger dataarray with 0 or 3 depending on the winner.

almostmeenal:
> > I was trying to do that with xarray.where and that didnt work out so well
> 
> Given that we only want the boolean result, you should be able to evaluate the condition directly: `(pp["home_points"] > pp["away_points"]) * 3` should return an interger dataarray with 0 or 3 depending on the winner.

tried it, didn't work. This gives you an array with all elements 3, for some reason element wise comparison isn't happening

OriolAbril:
> tried it, didn't work. This gives you an array with all elements 3, for some reason element wise comparison isn't happening

It works for me, are you still using where in addition to the code above?

almostmeenal:
> > tried it, didn't work. This gives you an array with all elements 3, for some reason element wise comparison isn't happening
> 
> It works for me, are you still using where in addition to the code above?

okayy my bad, thanks! I restarted my kernel and it worked, I guess it was some older version of the variable i used to store this array or something.

almostmeenal:
> Given that we only want the boolean result, you should be able to evaluate the condition directly: `(pp["home_points"] > pp["away_points"]) * 3` should return an interger dataarray with 0 or 3 depending on the winner.

```
pp["home_win"] = (pp["home_points"]>pp["away_points"])*3 + (pp["home_points"]==pp["away_points"])*2
pp["away_win"] = (pp["home_points"]<pp["away_points"])*3 + (pp["home_points"]==pp["away_points"])*2

groupby_sum_home = pp.home_win.groupby(const.home_team).sum().rename({"home_team": "team"})
groupby_sum_away = pp.away_win.groupby(const.away_team).sum().rename({"away_team": "team"})

pp["teamscores"] = (groupby_sum_home + groupby_sum_away)
pp["rank"] = pp["teamscores"].rank("team")
pp["rank"].where(pp["rank"]==1).groupby("team").count("draw")
```

okay this almost works, ranks are non integer sometimes, and maybe something is missing, I'll see that but maybe this is close, the results are weird though.

OriolAbril:
> ranks are non integer sometimes

It looks like dataarray.rank does not accept a `method` argument, and here we need to use the `min` one, not the default average. We'll need to use scipy instead. We can use it via `xr.apply_ufunc` in order to preserve coords and dims, I haven't tried but it should look like:

```
from scipy.stats import rankdata
xr.apply_ufunc(
    rankdata,
    - pp["teamscores"],  # comment on that later
    input_core_dims=[["team"]],
    output_core_dims=[["team"]],
    kwargs=dict(axis=-1, method="min")
)
```

The current rank, with the default average method averages the rank in case of ties. i.e. the following array `4, 12, 6, 12, 17` would become `1, 3.5, 2, 3.5, 5` the twelves would have rank 3 and 4 so they are averaged. We want `1, 3, 2, 3, 5` instead which is what happens in most competitions. This is the `min` method.

You'll see also that the lower number gets the higher rank! :warning: I added a `-` on the scores to take care of that, otherwise Italy will always win when it's clearly the worst of the 6 teams.

almostmeenal:
> from scipy.stats import rankdata
> xr.apply_ufunc(
>     rankdata,
>     - pp["teamscores"],  # comment on that later
>     input_core_dims=[["team"]],
>     output_core_dims=[["team"]],
>     kwargs=dict(axis=-1, method="min")
> )

Thankyou!! incorporated it in the latest commit, i guess its almost done now

OriolAbril:
Yeah, I think reproducing the initial results, even the `sim_table` is not far, I have a couple ideas on that. I also have one idea on how to extend the posterior predictive simulations and use the `size` argument of `sample_posterior_predictive` :grin:

almostmeenal:
> Yeah, I think reproducing the initial results, even the `sim_table` is not far, I have a couple ideas on that. I also have one idea on how to extend the posterior predictive simulations and use the `size` argument of `sample_posterior_predictive` 😁

okayy, I guess we can discuss them tonight?

almostmeenal:
this notebook is ready to be merged now!!

almostmeenal:
ig it works now

OriolAbril (approved):
The notebook looks amazing now!

All remaining comments are mostly styling. I'll merge after this last review iteration. Thanks for all the patience and for addressing all the comments.

almostmeenal:
@OriolAbril 
```
g = sns.FacetGrid(df_all, col="away_team", col_wrap=2, height=5)
g = g.map(plt.scatter, "year", "difference_non_abs").set_axis_labels("Year", "Score Difference")
```
gives an error: /Users/almostmeenal/anaconda3/envs/p/lib/python3.9/site-packages/seaborn/axisgrid.py:64: UserWarning: This figure was using constrained_layout==True, but that is incompatible with subplots_adjust and or tight_layout: setting constrained_layout==False.     self.fig.tight_layout(*args, **kwargs), i think this is happening after the arviz style, also i need to change the font size, is there something you suggest doing about this error?

almostmeenal:
@OriolAbril took care of all the comments, except the one with nqba-black, tried twice, apparently its adamant

OriolAbril:
> except the one with nqba-black, tried twice, apparently its adamant

wow, the reformat with the comments is terrible. Turn formatting of for this cell then, the style guide shows how: https://github.com/pymc-devs/pymc3/wiki/PyMC3-Jupyter-Notebook-Style-Guide

almostmeenal:
done

OriolAbril:
It looks like the notebook is not all executed now, maybe you saved and pushed before it finished?

almostmeenal:
> is somehow messing the theme,

the only problem is, without changing the font, the x labels are collapsing into each other and not at all legible

OriolAbril:
Try `g.fig.autofmt_xdate()` after `g.map...`

almostmeenal:
> Try `g.fig.autofmt_xdate()` after `g.map...`

thanks this worked!

OriolAbril:
> thanks this worked!

Great! I see the change in the first cell with the seaborn facetgrid, but not in the second, you should edit the 2nd cell (currently labeled In [13]) and then do a "Restart kernel & run all", all cells should be run sequentially to make sure everything works as expected.

almostmeenal:
> > thanks this worked!
> 
> Great! I see the change in the first cell with the seaborn facetgrid, but not in the second, you should edit the 2nd cell (currently labeled In [13]) and then do a "Restart kernel & run all", all cells should be run sequentially to make sure everything works as expected.

did

almostmeenal:
if this is done and ready to be merged then i can(?) finally remove my local branch, right now i'm constantly shifting between two branches

OriolAbril:
> >  then do a "Restart kernel & run all", all cells should be run sequentially to make sure everything works as expected.
>
> did

I still don't see all cells executed sequentially, there are some >30 in the middle of the notebook. I can rerun it myself too if you prefer

> if this is done and ready to be merged then i can(?) finally remove my local branch, right now i'm constantly shifting between two branches

Once the PR is merged github itself will show you a button to delete the branch, but switching between branches is good, you should get familiar with that and work on a single feature/change per PR and have each PR submitted from it's own branch.

almostmeenal:
> Once the PR is merged github itself will show you a button to delete the branch, but switching between branches is good, you should get familiar with that and work on a single feature/change per PR and have each PR submitted from it's own branch.

makes sense yeah

> I still don't see all cells executed sequentially, there are some >30 in the middle of the notebook.

ah yeah I didn't save the notebook after re run again my bad. it should be fixed in the last commit i made

OriolAbril:
Thanks for the PR!

Issue #57, opened 2021-03-29
log gaussian cox process


OriolAbril:
https://nbviewer.jupyter.org/github/pymc-devs/pymc-examples/blob/main/examples/case_studies/log-gaussian-cox-process.ipynb
Reviewers: @ckrapu

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
* use numpy [Generator](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator). See also https://numpy.org/doc/stable/reference/random/index.html?highlight=random%20sampling%20numpy%20random#quick-start

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### ArviZ related
* Use xarray and `from_pymc3_predictions` to filter nans and slice/reduce `intensity_samples`

### Exotic dependencies
None

### Computing requirements
Model takes roughly 5 mins to sample.

OriolAbril:
@ckrapu, feel free to decline. We are currently trying to update all the notebooks to make sure they follow pymc3 best practices and then have them run on pymc3 4.0+. We are tracking our progress in [this project](https://github.com/pymc-devs/pymc-examples/projects/1), it has a detailed description [here](https://github.com/pymc-devs/pymc-examples/blob/main/CONTRIBUTING.md#about-the-notebook-tracker-project). We submitted a project proposal to [outreachy](https://www.outreachy.org/apply/project-selection/) to have some extra help with that and therefore have a smooth transition from pymc3 v3 to pymc3 v4 with great docs on v4.

ckrapu:
No problem, I'm cleaning it up as we speak. This was a good chance for me to clean up the priors in the notebook so that there wouldn't be any need to filter out bad samples from the posterior for the gaussian process. As an aside, do you know how to suppress this warning? It's showing up 10-15 times per usage of `sample`.

```
/Users/v7k/anaconda3/envs/pymc3_dev/lib/python3.8/site-packages/aesara/graph/fg.py:500: UserWarning: Variable Elemwise{pow,no_inplace}.0 cannot be replaced; it isn't in the FunctionGraph
```

OriolAbril:
It looks like you are using aesara already which may be a bit too early, which pymc3 versions are you using?

[...]

PR #93, opened 2021-04-02
Revised LGCP notebook

ckrapu:
This PR updates the log Gaussian Cox process notebook to conform to best practices for documentation as documented in #57 . The following changes were made:

* The GP model formulation was changed from `ExpQuad` covariance kernel to `Matern52` and the diagonal variance prior was changed from `HalfCauchy` over the standard deviation to `InverseGamma` over the variance. This solved an issue with the earlier formulation in which some parameter draws corresponded to nonviable covariance matrices.
* As a consequence of the above, the code for filtering posterior samples for NaNs is no longer needed and was therefore removed.
* The usage of `np.random.randn` was replaced with the `rng.standard_normal` generator for random numbers.


OriolAbril:
Also note that you have to make sure to run the style checks: https://github.com/pymc-devs/pymc3/wiki/PyMC3-Jupyter-Notebook-Style-Guide

OriolAbril (approved):

Issue #46, opened 2021-03-24
sampler stats


OriolAbril:
file: https://github.com/pymc-devs/pymc-examples/blob/main/examples/diagnostics_and_criticism/sampler-stats.ipynb

## Required changes
The notebook should use InferenceData and ArviZ for plotting. Note that the names of the sampler stats in ArviZ are different, the naming convention for ArviZ can be found at https://arviz-devs.github.io/arviz/schema/schema.html#sample-stats, doc which should be linked too.

## Changes for discussion
I think it would be a good addition to add a plot_parallel as a quick way to visualize divergences or to link to the notebook on divergences.

almostmeenal:
that sounds like a good idea, I feel like adding more math-y info of those stats would be helpful, because in this notebook I think there's just trace.stat_names, which is sort of just documentation all over again, probably example notebooks should be doing more, I think what I'm trying to say is they can be somewhat pedagogical, I don't know if i make a lot of sense. But I'll be happy to work on it

canyon289:
Assigned you @almostmeenal. Based on our early conversations regarding GSOC and your interests this seems like a great one to both help you work on the types of OSS youd like to, as well as to keep building up your experience for a strong application (for any project you apply to)!

almostmeenal:
thanks! I'll get to it soon, yes this does look like a right fit.

PR #94, opened 2021-04-02
Samplerstats(WIP)

almostmeenal:
This notebook uses InferenceData and ArviZ for plotting.
Changed sampler_stats description according to - https://arviz-devs.github.io/arviz/schema/schema.html#sample-stats

Addresses issue #46 

Still working on:

- [ ] parallel plot idea?
- [ ] pedagogical info about sampler stats
- [ ] possibly expanding and adding a few more plots about sampler stats(try and see if more arviz plots can be incorporated - like dist_plot etc)

PR #95, opened 2021-04-02
sampler stats

almostmeenal:
This notebook uses InferenceData and ArviZ for plotting. Explains multi-step sampler_stats as well.
Changed sampler_stats description according to - https://arviz-devs.github.io/arviz/schema/schema.html#sample-stats

Addresses issue #46


almostmeenal:
deleted PR Samplerstats(WIP) #94 and created this, will take care of comments made there and make a commit, sorry about the inconvenience, i tried resetting to an initial commit and that messed up the previous PR so started it again

almostmeenal:
I tried `trace.sample_stats["tree_depth"].plot(row="chain", ls="none", marker=".", alpha=.3)
`
but I think because of the number of points or something, it still resembles a line plot, visually looks way better though.

image

OriolAbril:
> I tried trace.sample_stats["tree_depth"].plot(row="chain", ls="none", marker=".", alpha=.3)
but I think because of the number of points or something, it still resembles a line plot, visually looks way better though.

Let's go with `col` instead of hue then, but keep the alpha. `tree_depth` is an integer, and it has one value per draw, as there isn't much variation, it generates these visually looking flat lines, but you can see that on 1 there are some stray points. I think the scatterplot with col will look good enough.

The plot step_size bar is the one I find useless, we should just remove it and be done with it.

The comment about using `plot_posterior` was for the histograms for the acceptance rates, that you have already updated, not for this step size plot.

examples/diagnostics_and_criticism/sampler-stats.ipynb
```diff
@@ -14,29 +14,40 @@
   },
```
OriolAbril, line 14: <p>Sorry, it should be <code>list(trace.sample_stats.data_vars)</code> we only want to see the names here</p><p> </p><p>EDIT: actually scratch that, we should use that below with the multiple samples, but show the html repr of <code>trace.sample_stats</code>  here so we can then comment on both variable names _and attributes!_ There is nothing on attributes because there did not use to be here. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

OriolAbril, line 14: <p>You can delete this plot along with the markdown cell explaining it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

OriolAbril, line 14: <p>This should be "find out if there have been divergent transitions and how many"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

OriolAbril, line 14: <p>Add an explanation under the sum here saying that if they do have divergences, they should go read https://docs.pymc.io/pymc-examples/examples/diagnostics_and_criticism/Diagnosing_biased_Inference_with_Divergences.html</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

OriolAbril, line 14: <p>I'd delete this where subsetting as it's not really useful here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

OriolAbril, line 14: <p>dims and coords seem to have been ignored, could be a bug in ArviZ. I'll look into that. Do change this for the plot anyway though, even if at least for now the labels are wrong. Using <code>plot_posterior</code>  like we did above should work here too.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

almostmeenal: <p>so for the first sampling we go for, <code>trace.sample_stats</code>, followed by markdown description?</p>

almostmeenal: <p>okayy</p>

OriolAbril: ArviZ ignores the dimensions for sample stats: https://github.com/arviz-devs/arviz/blob/main/arviz/data/io_pymc3.py#L323 it may be to avoid name collisions though.

examples/diagnostics_and_criticism/sampler-stats.ipynb
```diff
@@ -21,22 +21,34 @@
      "name": "stdout",
```
OriolAbril, line 21: <p>I think xarray is not used explicitly anymore so we can remove the import</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

OriolAbril, line 21: <p>Note that not all the parameters defined in ArviZ schema are present nor all the variables in <code>trace.sample_stats</code>  are in ArviZ schema, some are specific to PyMC3. We should not replace the original definitions by the descriptions in arviz shcema but combine them. </p><p> </p><p>There are also some new variables that have no explanation in the old notebook, I think these are related to the extra info @michaelosthege added to the trace report.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

OriolAbril, line 21: <p>This paragraph about the name clashes also has to go. In this cell there should instead an explanation on how InferenceData stores extra info like the date, versions used and sampling time as attributes</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

OriolAbril, line 21: <p>The note on <code>model_logp</code>  has to be updated</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

OriolAbril, line 21: <p>I would remove this cell, seeing values for <code>accept</code>  doesn't provide any info, the plot alone is enough</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

OriolAbril, line 21: <p>I'd use kind="hist" here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

examples/diagnostics_and_criticism/sampler-stats.ipynb
almostmeenal: <p>do you think that could use some fixing, maybe an informative  warning or error could be given during a name collision. Because in real time models and stuff diagnostics running well, following general practices of arviz/pymc3 seems to be actually very important</p>

examples/diagnostics_and_criticism/sampler-stats.ipynb
almostmeenal: <p>um that would be weird, combining them I mean, because some variables are like -</p><p> </p><p>step_size_nom in arviz and step_size_bar in pymc3 and so which one should we go for? should we put all the variables, if thats what you meant? also i think the total number of variables and names should be consistent in pymc3 and arviz if we are using arviz for sampler_stats what we can do is, just not use some of the variables in each library when theyre not relevant, but they should be defined in both the libraries? does that make  any sense</p>

almostmeenal: um this comes out bad:
image

almostmeenal: <p>multiple samplers doesnt have a variable model_logp, so i removed the cell there, and lp is defined for the first sampling case</p>

examples/diagnostics_and_criticism/sampler-stats.ipynb
```diff
@@ -21,11 +21,12 @@
...
```
almostmeenal, line 21: <p>@OriolAbril I added bins here in favor of better visualization, because bins in auto mode, had x_ticks from 0 to 400, which made it impossible to visualize the accept 1 hist plot</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

almostmeenal, line 21: <p><a href="https://github.com/OriolAbril" target="_blank">@OriolAbril</a> I added bins here in favor of better visualization, because bins in auto mode, had x_ticks from 0 to 400, which made it impossible to visualize the accept 1 hist plot</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

OriolAbril: <p>I will look into that. We should not use <code>bins</code>  because now it visually looks better, but the fact that it originally went up to 400-600 is because there are values as high as that in <code>accept</code>,  no idea why though, I think it should be a probability.</p>

OriolAbril, line 21: <p>Here keep <code>trace.sample_stats</code>  so readers can see both variable names and attributes all at once</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

OriolAbril, line 21: <p>This list should contain only the variables that we see in <code>trace.sample_stats</code>.  Therefore it will contain most varibles explained in the ArviZ schema, but not all, <code>step_size_nom</code>  or <code>int_time</code>  are not present in pymc3 but are defined in arviz because other PPLs like Stan or Turing (in Julia) do have them. It will also contain variables like <code>step_size_bar</code>  or <code>perf_counter_diff</code>  which are present here but not on ArviZ schema. Right now there are several variables which are currently replicated because they use both the internal pymc3 name and the arviz one. This <a href="in Julia" target="_blank">comment</a> might help a bit, but the important thing is to have a bullet point per variable in <code>trace.sample_stats</code>.  Nothing more but nothing less either.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

OriolAbril, line 21: <p>This paragraph on name clashes should be deleted. Now variables in the posterior are available as <code>trace.posterior["var_name"]</code>  and sample stats as <code>trace.sample_stats["stats_name"]</code>,  there is no <code>trace["var_name"]</code>  option.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

examples/diagnostics_and_criticism/sampler-stats.ipynb
OriolAbril: Yes, we could fix that in ArviZ and have it use coords and dims also in the sample stats, either with a try except or checking things more carefully. But this is very low priority, compound steps are not that common and we have to make sure to implement that in a way that it never backfires.

examples/diagnostics_and_criticism/sampler-stats.ipynb
almostmeenal, line 21: <p>over here, I also get this notif:</p><p> </p><pre class="ql-syntax" spellcheck="false">The number of effective samples is smaller than 10% for some parameters.
</pre><p> </p><p>I checked <a href="https://discourse.pymc.io/t/the-number-of-effective-samples-is-smaller-than-25-for-some-parameters/1050/3" target="_blank">this page</a> for a similar thing, should we be addressing this? mentioning something about what to do if a similar thing comes up?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

almostmeenal: that comment is not available, i probably do not have access to that link. Anyway I tried to make the changes after having a look at NUTS stats

almostmeenal:
made some changes(experimental)

examples/diagnostics_and_criticism/sampler-stats.ipynb
OriolAbril: No need to worry about that here

OriolAbril: Sorry,  link got corrupted when pasting,  here it is again https://github.com/arviz-devs/arviz/pull/1063#issuecomment-589771200 (I still have not checked reviewnb, currently on mobile at the train)

almostmeenal: @OriolAbril no problem at all! also I just checked and the items ive added are the exact same, I went through the code for sample_stats_to_array [here ](https://github.com/arviz-devs/arviz/blob/main/arviz/data/io_pymc3.py#L323 ), that clarified it for me. thanks anyway this looks like a helpful comment.

this is from the current commit on the notebook, in case you wanted to see:
image

OriolAbril:
@michaelosthege can you explain what are perf_counter_diff , process_time_diff and perf_counter_start? Or point us to an issue or PR about them?

almostmeenal:
> @michaelosthege can you explain what are perf_counter_diff , process_time_diff and perf_counter_start? Or point us to an issue or PR about them?

hi, maybe this comment wasn't notified or something because it was in the reviewnb thing, 
https://github.com/pymc-devs/pymc-examples/pull/95#discussion_r606847518 , but I found these descriptions in the NUTS sampler documentation

almostmeenal:
i think it still doesn't open in the mobile app, so im copying the comment here:

@OriolAbril no problem at all! also I just checked and the items ive added are the exact same, I went through the code for sample_stats_to_array [here ](https://github.com/arviz-devs/arviz/blob/main/arviz/data/io_pymc3.py#L323 ), that clarified it for me. thanks anyway this looks like a helpful comment.

this is from the current commit on the notebook, in case you wanted to see(only tune needs to go i think)
image

almostmeenal:
[here](https://docs.pymc.io/api/inference.html#module-pymc3.step_methods.hmc.nuts) is the relevant documentation

OriolAbril:
Yeah, I am seeing it now, sorry about that. Thanks for looking into this.

OriolAbril:
You are right that now all the relevant variables are present, only `tune` has to be removed.

Regaring the `accept` variable from metropolis, I am not sure it "deserves" so much analysis, but I still don't understand what it is

almostmeenal:
> You are right that now all the relevant variables are present, only `tune` has to be removed.
> 
> Regaring the `accept` variable from metropolis, I am not sure it "deserves" so much analysis, but I still don't understand what it is. The analysis is great though

yeah that is somewhat weird, all i could find in metropolis.py was:       
 `accept = self.delta_logp(q, q0)`

Also tune is a stat in metropolis.py but I don't see it in sample_stats, so I'm a little confused about whether it should be needed or not. I think it got removed because of the compound step thing, if it's there in a single model sampling for metropolis dist. then we might need it?

almostmeenal:
I looked at the original notebook once again, and possibly the reason why they chose to show accept, was just that it was common in both the samplers, which might not be a good enough reason for an elaborate plot which also has outliers. I might be entirely wrong here, I'm just speculating.

OriolAbril:
> and yeah that is somewhat weird, all i could find in metropolis.py was:
accept = self.delta_logp(q, q0)

I took a look at that, now it make sense. In metropolis one generates a proposal `q` and then compares it's probability to the current sample `q0` with `alpha = P(q) / P(q0)`. If `alpha > 1`, that is, `q` has higher probability than `q0` the proposal is always accepted, otherwise `alpha` is in the range `0, 1` so sample from the uniform distribution `u` is generated, if `u>alpha` the proposal is rejected, accepted otherwise. `accept` is `alpha` which is also sometimes called `acceptance ratio`

> I looked at the original notebook once again, and possibly the reason why they chose to show accept, was just that it was common in both the samplers, which might not be a good enough reason for an elaborate plot which also has outliers. I might be entirely wrong here, I'm just speculating.

yes, this sounds right, our goal here should probably be to show that when using a compound step, the common variables in sample_stats will get an extra dimension, one per sampler.

almostmeenal:
> I took a look at that, now it make sense. In metropolis one generates a proposal `q` and then compares it's probability to the current sample `q0` with `alpha = P(q) / P(q0)`. If `alpha > 1`, that is, `q` has higher probability than `q0` the proposal is always accepted, otherwise `alpha` is in the range `0, 1` so sample from the uniform distribution `u` is generated, if `u>alpha` the proposal is rejected, accepted otherwise. `accept` is `alpha` which is also sometimes called `acceptance ratio`

yeah, that makes sense, thanks for explaining, i just started reading up about it and noticed acceptance rate was important for metropolis hastings.

> yes, this sounds right, our goal here should probably be to show that when using a compound step, the common variables in sample_stats will get an extra dimension, one per sampler.

In that case, the last plot is unnecessary, but possibly showing the range of accept on both the samplers, would make some sense?

OriolAbril:
> acceptance rate was important for metropolis hastings.

In metropolis it's `acceptance ratio` which is very different from the `acceptance rate` in hmc/nuts

> In that case, the last plot is unnecessary, but possibly showing the range of accept on both the samplers, would make some sense?

Yes, I think that showing how to work with sample stats that have multiple steps is useful and won't be explained anywhere else probably

almostmeenal:
> In metropolis it's `acceptance ratio` which is very different from the `acceptance rate` in hmc/nuts

yes, my bad, ive been looking at nuts doc. over and over for the stats thing sort of fixated in my head.



> Yes, I think that showing how to work with sample stats that have multiple steps is useful and won't be explained anywhere else probably

yes we can do that, also I think a detailed explanation should be added here as well - [compound step notebook](https://github.com/pymc-devs/pymc-examples/blob/a093bcc5aa828e9317bdf9f936f968f76acdb4f7/examples/pymc3_howto/sampling_compound_step.ipynb)

OriolAbril:
> yes we can do that, also I think a detailed explanation could be added here as well - compound step notebook

good catch, I _think_ I haven't gotten to create the issue for that notebook, I don't remember ever reading it until now.

almostmeenal:
> > yes we can do that, also I think a detailed explanation could be added here as well - compound step notebook
> 
> good catch, I _think_ I haven't gotten to create the issue for that notebook, I don't remember ever reading it until now.

thanks! I haven't come across an issue on that, if you want I can make one, or I can directly make a draft PR too, whichever seems better?

OriolAbril:
I will create an issue now

almostmeenal:
> I will create an issue now

sure

examples/diagnostics_and_criticism/sampler-stats.ipynb
OriolAbril, line 21: <p>This should be clear in that those names are internal (<code>return_inferencedata=True</code>  will become the default so in the long run those probably won't be accessible at all).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

OriolAbril, line 21: <p>From the introduction before the bullet points, I understand that all these variables are part of the ArviZ schema and not all are. I would use something like "Some of the sample statistics used by NUTS are renamed when converting to <code>InferenceData</code>  to follow [ArviZ's naming convention], whereas some others are specific to PyMC3 and keep their internal PyMC3 name in the resulting <code>InferenceData</code>  object.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

OriolAbril, line 21: <p><code>mu1</code>  and <code>mu2</code>  should not have <code>dims="step"</code>,  this is having them both be of shape <code>(2,)</code>.  <code>mu1</code>  should be a scalar and <code>mu2</code>  should have shape <code>(10,)</code>  (you can create and <code>obs</code>  dim to use in <code>mu2</code>),  only <code>accept</code>  has this <code>step</code>  dimension (which is ignored but could be fixed and if so the plot will update the labels automatically).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

OriolAbril, line 21: <p>I would not call them outliers, those are simply jumps from regions of low probability to regions of much higher probability, they are bound to happen at some point and not completely unexpected.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

OriolAbril, line 21: <p>I would probably use <code>trace.sample_stats["accept"].max("draw") - trace.sample_stats["accept"].min("draw")</code>  <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/95/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

OriolAbril:
can you restart and run all on the notebook? This should get rid of this warning in the last cell due to having executed it twice:

> The watermark extension is already loaded. To reload it, use:
  %reload_ext watermark

almostmeenal:
@OriolAbril done, it can be merged now

PR #97, opened 2021-04-04
Check execution order

MarcoGorelli:
## Description

Following up on a suggestion by @OriolAbril :

> How difficult would it be to add a pre-commit check that ensures notebooks have been executed sequentially? Or an extra ci check with another tool?

Indeed, there are some notebooks which have cells which are out-of-order! So, for the time being, am excluding them from the check, then they can be re-run and put back in to the check later


OriolAbril (approved):
Thanks!

PR #141, opened 2021-04-05
Modify compound steps notebook

almostmeenal:
This PR addresses issue #113 

- Adds Arviz Inference Data
- Uses Coords and Dims
- Adds an arviz plot of sample stats to understand them for compound steps, as it is not explicitly explained anywhere else
- Info on sample stats variables common between samplers, noticed this on #95 but this seems to be a more relevant place for this explanation


almostmeenal:
accidental, fixing and rebasing.

almostmeenal:
accidental, fixing and rebasing

almostmeenal:
finally fixed this git rebase thing oof, excuse the bad open/close history ;-;

almostmeenal:
@OriolAbril I considered explaining the two variable part of compound sampling over here, and just mentioning it along with showing the outliers in the sample_stats notebook. Does that make sense?

examples/pymc3_howto/sampling_compound_step.ipynb
```diff
@@ -12,10 +12,24 @@
     "# Compound Steps in Sampling\n",
```
OriolAbril, line 12: <p>"What happens to <s>variables</s> that occur in multiple <s>samplers</s>? " it should be sample statistics and step methods, actual variables can't be sampled by two step methods at once. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/141/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

OriolAbril, line 12: <p>I would not list the available step methods here as it will get outdated in no time. I would also move that to the compound steps by default secion.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/141/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

OriolAbril, line 12: <p>az.style.use should go in the 2nd cell, somethimes matplotlib doesn't load the style correctly if it's imported in the same cell the style is set (matplotlib is imported by arviz)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/141/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

OriolAbril, line 12: <p>return_inferencedata=True</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/141/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

OriolAbril, line 12: <p>I would keep the original code because it has 2 advantages. 1st is that it shows how one can sample the same model as many times as they want with different samplers, iterations or whatever. 2nd is that this is an actual model with observed variables, the mu1, mu2 example only samples from free variables, it doen't sample from a posterior. We can still use coords and dims by using <code>idata_kwargs</code>  (in which case we don't need to use coords when defining the model)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/141/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

OriolAbril, line 12: <p>I would sample from the model using both orders of step methods and add the sample stats section here, showing how each step method gets it's own accept (here accept being log p_updated - log p_current, it is already defined here) and how inverting them affects that. Note that that actually forces us to use idata_kwargs instead of coords at model creation, because even if the dims don't change, the coordinates do.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/141/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

almostmeenal: <p>I tried keeping the original code and for some reason that didn't give me two values for "accept", so I was somewhat confused</p>

almostmeenal: <p>I'll get back to this one, once I get the existing code working </p>

almostmeenal:
@OriolAbril I've made changes except the last two comments, can you have a look and tell me if i missed something, was idata_kwargs supposed to be given once, or during every consecutive event of sampling? (else I'll get back to it tomo and go over again)
(check latest commit for context)

examples/pymc3_howto/sampling_compound_step.ipynb
OriolAbril: I think the original code was not sampling with the custom steps defined, only printing them to show which order they define, so trace probably referred to the trace from the other model, sampled combining nuts and metropolis instead of combining two metropolis.

almostmeenal: in the latest commit, i sampled after defining steps

OriolAbril: Took a look at the code, it's probably due to BinaryGibbsMetropolis not having accept unline BinaryMetropolis, so there is no repeated name on which to concatenate the accept stat

almostmeenal: so I should go for binarymetropolis and metropolis instead of binarygibbs and metropolis right?

OriolAbril: I would use `BinaryMetropolis`. I am not sure how used are compound methods, but all the order and other explanations done with BinaryGibbsMetropolis will still be relevant, we will only be adding the repreated stats component

almostmeenal: yes, it worked like a charm, I'll push the latest commit, I think it looks almost ready now

examples/pymc3_howto/sampling_compound_step.ipynb
```diff
@@ -12,10 +12,11 @@
...
```
OriolAbril, line 12: <p>I would add that comment as a markdown cell and not print the whole dataarray</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/141/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

OriolAbril, line 12: <p>Add theano and xarray here, I added how to do that to the style guide: https://github.com/pymc-devs/pymc3/wiki/PyMC3-Jupyter-Notebook-Style-Guide</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/141/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='32'/>

almostmeenal: <p>does printing one chain work?</p>

almostmeenal: <p><code>trace1.sample_stats["accept"].sel(chain=1).values</code></p><p> </p><p>I'm going for this because it helps actually notice the two sampler step accept thing and this doesnt take up much space</p>

almostmeenal:
@OriolAbril thanks for the suggestions! everything's done

OriolAbril (approved):

PR #142, opened 2021-04-06
Moderation analysis example

drbenvincent:
Hi all
Here's a pull request adding an example notebook covering moderation analysis.  I wasn't quite sure on the policy of where to store an embedded image, so I've just included one in the directory that the notebook is in. Just let me know if you need any changes made.


ricardoV94:
This is really neat! The visualizations with the moderator variable really speak for themselves. 

I left some minor comments if you want to address them :)

drbenvincent:
Thanks @ricardoV94, I should be able to deal with these points soon.

fonnesbeck:
This is great, thank you. The data import should probably use `pm.get_data`, rather than a relative path string.

drbenvincent:
Excuse the slight delay... I had to get a couple of papers submitted. I believe I've not implemented all the changes asked for. The last changes were to provide `idata_kwargs` when sampling, which enabled use of xarray, which has meant I could kill a utility function to extract samples which has lead to smoother plotting code. Let me know if there's anything else :)

drbenvincent:
Thanks for those points @OriolAbril. Unless anyone has any further things, I think that might be it. 🤞🏼

OriolAbril (approved):
Thanks for the PR! I'll happily add this notebook directly to the "Best Practices" column

Issue #99, opened 2021-04-05
dependent density regression


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/mixture_models/dependent_density_regression.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

chiral-carbon:
this notebook is updated except for a minor thing in the beginning where it does not use `numpy Generator` instead of `numpy.random`. can I open a PR for this?

OriolAbril:
It should also add the posterior predictive samples to the InferenceData and update the plotting in the last cell. `low, high = np.percentile(pp_trace["obs"], [2.5, 97.5], axis=0)` for example could use xarray and specify which dimensions are reduced explicitly.

Note that unlike most notebooks, this one is not uncategorized within the notebook tracker project but it's triaged to the "to do" indicating that both changes related to arviz and to general updates are needed

PR #146, opened 2021-04-24
Use coords / xarray magic in dependent density regression

MarcoGorelli:
xref #14 

Is there a way to avoid having to do
```python
beta = pm.Normal("beta", 0.0, 5.0, shape=(1, K))
```
?
If there was `shape=K`, I could put `dims='K'`, but then
```python
pm.math.dot(x, beta)
```
wouldn't work.

Any hints?

EDIT (Oriol): Addresses https://github.com/pymc-devs/pymc-examples/issues/99


examples/mixture_models/dependent_density_regression.ipynb
```diff
@@ -2,16 +2,7 @@
  "cells": [
```
OriolAbril, line 2: <p>You can use <code>"one": [1]</code>  in <code>coords</code>  dict, or any other name/coord value you may prefer, it just has to be there and have length one to act as the <code>1</code>  in shape.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/146/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

MarcoGorelli: <p>so simple, thanks!</p>

MarcoGorelli, line 2: <p>so much more readable with xarray magic...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/146/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

OriolAbril: <p>:) </p><p> </p><p>Updating the examples is being an amazing resource for me to both see the potential of xarray as well as where is more help and work needed. I don't need convincing on xarray usage, but there are still things where it surprises me. </p><p> </p><p>As a side not, I think somehting like <code>max_mixture_weights["K"].values</code>  will be equivalent to<code>max_mixture_weights.coords.to_index()</code> . The distinction between coord and variable is still something I don't grok yet, especially because of non-indexing coords, but it is cool to be able to treat coords as variables in many cases </p>

fonnesbeck:
Thanks, @MarcoGorelli

Issue #91, opened 2021-03-30
generalized linear models


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/generalized_linear_models/GLM.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

KavyaJaiswal:
I want to work on this.

OriolAbril:
There seem to be issues when trying to run the notebook. It doesn't work on v3, not sure why, I don't know how GLM module works. If someone does great, otherwise it may not be worth fixing if it's a bug, we probably can update this to use bambi directly.

almostmeenal:
hey @KavyaJaiswal, i guess since you'll be working on outreachy, if you've left this issue i can take it up!!

KavyaJaiswal:
Sure, all the best!!

OriolAbril:
@almostmeenal do you plan to start on this right away? 

I think the best way forward is to merge @KavyaJaiswal's PR which already covers a lot of important updates so that you can then fix the second half of the notebook. Now that [Bambi 0.5 is out](https://discourse.pymc.io/t/bambi-0-5-0-is-out/7463), once the current PR is merged most of the work will be moving from pymc3.gml module to bambi. Sounds good?

almostmeenal:
okayy!! that sounds good, you can merge the PR and then tomorrow i'll have a look at what to do next?

OriolAbril:
General note that while I would not consider having the notebook "broken" a blocker, it would be great to port it to bambi before 3.11.3 is released and therefore the website is updated. I can also ask bambi devs for help if this timeline is too short

almostmeenal:
hey, okay yes that makes sense. can you give me a deadline (for personal organization) deadlines help me avoid procrastinating, then i'll get it done!!!

OriolAbril:
[...]

PR #148, opened 2021-04-29
Update first half of generalized linear models to "Best Practices"

KavyaJaiswal:
## Description

https://github.com/pymc-devs/pymc-examples/blob/main/examples/generalized_linear_models/GLM.ipynb

Addresses issue #91 and aims to advance the notebook to the 'Best Practices' State. 

The new code uses the new numpy generator instead of the numpy random. The code also has an ArviZ related addition of return_inferencedata=True

The PR has a try... except clause in the code. It will ensure that users who have cloned pymc-examples repo will read their local copy of the data while also downloading the data from github for those who don't have a local copy. 
```
try:
    sat_data = pd.read_csv("../data/Guber1999data.txt")
except:
    sat_data = pd.read_csv(pm.get_data("Guber1999data.txt"))
```

This PR also addresses the error message in the Hierarchical GLM section of the code. The error in the code is shown in the screenshots attached below. The fix for this error can be discussed and worked upon.

Error1
Error2
Error3


KavyaJaiswal:
I have added all the required changes.

OriolAbril (commented):
Two general comments. 

First one is to update the PR title so when we merge it and it becomes a commit message, it will be informative of the changes, to do so it needs to mention the notebook that was modified. Most notebooks will be updated to use the new generator, so without that we'd end up with ~70 identical commit messages. 

Second one is to remember to run `pre-commit` to ensure that CI passes. There is some guidance on that at https://github.com/pymc-devs/pymc3/wiki/PyMC3-Python-Code-Style

KavyaJaiswal:
Updated the Pull Request with the suggested changes,

Issue #81, opened 2021-03-30
GLM linear


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/generalized_linear_models/GLM-linear.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

chiral-carbon:
Hello! I commented under issue #58  too but this issue is also one I'd like to take up and attempt to make the suggested improvements.

OriolAbril:
I'll assign this to you too,  as PRs are only on a per notebook basis here, it's no problem to work on 2-3 notebooks simultaneously

chiral-carbon:
great, thank you!

chiral-carbon:
hello! had a plots related doubt. apart from using `figzise` as kwargs in `plot_trace` in cell 7 (using ArviZ), is there any other change that should be done? it's a very minor change so I wanted to confirm in case I might have misunderstood. 

the plot code looks like this currently:
'az.plot_trace(trace, figsize=(11,7));`

is this sufficient?

OriolAbril:
Yes, the notebook is short and concise and there aren't many things to update

chiral-carbon:
okay, got it. thanks.

OriolAbril:
Needs to be updated to use bambi instead of glm module

drbenvincent:
Link to notebook broken. Does this notebook exist any more?

OriolAbril:
This was moved to https://docs.pymc.io/en/latest/learn/core_notebooks/GLM_linear.html. It has already been updated to v4 and is part of the subset being auto-executed every time we build docs

PR #149, opened 2021-04-29
Update GLM-linear to use numpy.Generator and ArviZ

chiral-carbon:
Addresses issue #81 and aims to make the following changes:

### General Updates
- Uses numpy generator

### ArviZ related
- Uses `figsize`  as kwarg to run `plot_trace` function
 
-----------------------------------------------------------------------------
### Pending
- Checking why the manual model runs slower


examples/generalized_linear_models/GLM-linear.ipynb
```diff
@@ -103,9 +103,6 @@
    "metadata": {},
```
OriolAbril, line 103: <p>Even though we only use it once, I would keep the rng creation in the 2nd cell plus consequent usage of the <code>rng</code>  object, like the example in https://github.com/pymc-devs/pymc3/wiki/PyMC3-Jupyter-Notebook-Style-Guide</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/149/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

OriolAbril, line 103: <p>Can you also add the versions of theano and xarray? I added how to in the last point at https://github.com/pymc-devs/pymc3/wiki/PyMC3-Jupyter-Notebook-Style-Guide</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/149/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

OriolAbril:
Looks good, thanks! 

I have two extra general comments in addition to the ones above. The first one is making sure that you have executed the notebook sequentially after editing the code to make sure everything runs correctly, the second is to modify the PR title to something more descriptive _in the context of this repo_, sooner or later all notebooks will be updated for arviz, numpy generator and so on, so finding commits with messages about ArviZ would be difficult, some options that come to mind "update glm-linear to v3 best practices" or "update glm-linear to use numpy.Generator and ArviZ"

examples/generalized_linear_models/GLM-linear.ipynb
chiral-carbon: Sure, I'll add that.

chiral-carbon: xarray is not being used in this particular notebook so would it be okay to add to it in metadata?

chiral-carbon:
thanks for the suggestions, I'll push these changes. 
Regarding running the notebook, I did run it for testing if the code changes work. But in this particular commit, I have only made the code changes without running the notebook, the reason being that for some reason the manual model seems to run faster for me and the glm.from_formula takes more time. I wanted to ask if you might know what the reason for it could be - is there some dependency that I might be missing or something else because of which I get this result?

OriolAbril:
> the reason being that for some reason the manual model seems to run faster for me and the glm.from_formula takes more time. I wanted to ask if you might know what the reason for it could be - is there some dependency that I might be missing or something else because of which I get this result?

I don't really know, if the performance difference is not systematic you can then ignore the comment on the issue. It is also not worth to go into the source code of the glm module as it will probably be deprecated and replaced by [bambi](https://github.com/bambinos/bambi) (as an external dependency) in `v4`

examples/generalized_linear_models/GLM-linear.ipynb
OriolAbril: It looks like it's not being used but it is being used by ArviZ under the hood, like pymc3 uses theano under the hood even if we don't import or call it explicitly.

chiral-carbon:
Okay, so if I push a new commit with the suggested changes (adding xarray version and keeping `default _rng()` in `cell 2`) should I run the notebook too or not? the notebook will execute, but as I mentioned the output in the model cells would change, and since it may be an isolated issue in my case since we don't know the cause, the output might reflect an inaccurate performance of the model if I run it?

OriolAbril:
Performance will always vary between computers, the runtimes in the docs are at most orientative of the order of magnitude, we should always execute the notebooks and push the changes when modified

chiral-carbon:
understood. thanks for clarifying.

chiral-carbon:
so I tried executing ```%watermark -n -u -v -iv -w (-p theano xarray)``` and it returned a usage error:

```UsageError: unrecognized arguments: (-p theano xarray)```

I tried several things and realized that passing -p PACKAGES with a single package works, but it somehow fails with multiple arguments for me. 

this seemed to work:

```%load_ext watermark

%watermark -n -u -v -iv -p theano 
%watermark -p xarray -w``` 

and it outputs:

```The watermark extension is already loaded. To reload it, use:
  %reload_ext watermark
Last updated: Sat May 01 2021

Python implementation: CPython
Python version       : 3.8.5
IPython version      : 7.13.0

theano: 1.1.2

pymc3     : 3.11.2
numpy     : 1.18.5
arviz     : 0.11.2
matplotlib: 3.3.4

xarray: 0.17.0

Watermark: 2.2.0
```

I wanted your input as to whether you also get this error and if the correction I provided will be acceptable since passing `(-p theano xarray)` is not working in my notebook.

OriolAbril:
I think the error comes from the parenthesis. They are in the guide to indicate that this extra args may not be necessary, many notebooks import theano and xarray explicitly and their versions are listed automatically

chiral-carbon:
okay. running `%watermark -n -u -v -iv -w -p theano xarray` however also returns a usage error, as it doesn't recognize the second arg passed (in this case xarray). The error in this case is: ```UsageError: unrecognized arguments: xarray```.

Should I push the commit by breaking the `%watermark` statement into 2 statements to pass the flags as I have mentioned above? It doesn't align with the style guide but prints the package versions as desired.

OriolAbril:
My bad,  I assumed spaces would work without looking carefully into it. It looks like the separation between theano and xarray has to be a comma instead.

Source: https://github.com/rasbt/watermark/blob/master/watermark/watermark.py#L180

chiral-carbon:
great, got it!

OriolAbril (approved):

Issue #86, opened 2021-03-30
GLM poisson


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/generalized_linear_models/GLM-poisson-regression.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

jessicakzhang:
Hi! I'd like to try working on this.

OriolAbril:
That would be great @jessicakzhang! I have added a couple of suggestions based on a quick look over the notebook, I'll review more carefully once you submit a PR. 

Let us know if you have any doubt while working on this

chiral-carbon:
Hi @jessicakzhang , are you still working on this issue? @OriolAbril would it be okay if I were to submit a PR for this issue, considering the fact that this issue has already been assigned?

OriolAbril:
Hi, yes, as it has been more that two weeks with no activity, as indicated in the contributing guide, I'll assign the issue to you so you can submit a PR.

chiral-carbon:
thanks a lot!

chiral-carbon:
@OriolAbril had a doubt regarding updating `np.exp(np.mean())` to `np.mean(np.exp())` in cells 15 and 19. as far as I understood, `az.summary` returns certain statistics of which mean is one, and we are computing `np.exp()` for this summary dataframe. 
should this be changed to compute `np.mean()` of `az.summary`? I am unclear as to exactly where we should be applying `np.exp()` here.

OriolAbril:
I used mean as a placeholder, `summary` is acting as mean (as well as acting as hdi). Exponentiating should come first, then calling summary on the exponentiated data, not the other way around. 

I think the default stats is good enough and it's simple, there is no need to overly complicate the notebook only to exclude `sd` from summary.

chiral-carbon:
[...]

PR #154, opened 2021-05-06
Update GLM-poisson-regression to best practices state

chiral-carbon:
Addresses issue #86 and aims to advance it to best practices state.

### General Updates
- Uses numpy.Generator
- Modifies cells 15 and 19 to compute exponent before calling `az.summary`

### ArviZ Related
- Uses arg `kind=stats` rather than manually subsetting columns for stats in cells 15 and 19 to display summary


OriolAbril (commented):
Thanks for the PR, I think it can be merged after updating the watermark

OriolAbril (approved):

Issue #90, opened 2021-03-30
GLM truncated & censored


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/generalized_linear_models/GLM-truncated-censored-regression.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

drbenvincent:
I had this example notebook accepted a few months ago - never made it live on the website. I think it just needs some small things done (i.e. from the [Notebook updates overview](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)) to follow best practice. Happy for you to assign me to this.

drbenvincent:
Just checking the procedure... I should make a pull request when I'm done, and reference this issue?

OriolAbril:
Yes, when you open the PR link this issue so it will keep track of all the evolution of the notebook. Once merged I'll triage it to "best practices" and github will add the "oriol moved this issue to best practices in notebook tracker", then someone will make a PR for v4 also linkong here...

chiral-carbon:
can I take this up?

drbenvincent:
This is now done here in a notebook that has been merged https://github.com/pymc-devs/pymc-examples/blob/main/examples/generalized_linear_models/GLM-truncated-censored-regression.ipynb
 I think we can close this issue now?

OriolAbril:
[...]

PR #159, opened 2021-05-11
Truncated/Censored regression example

drbenvincent:
Updating this truncated/censored regression example notebook to follow best practice, as fas as I can tell. See #90. Let me know if there's anything else that needs to be done.


OriolAbril (commented):
Changes look good, thanks! The issue is that it looks like you took the original branch where you added the example and this resulted in merge conflicts. These need to be fixed before we can merge.

drbenvincent:
> Changes look good, thanks! The issue is that it looks like you took the original branch where you added the example and this resulted in merge conflicts. These need to be fixed before we can merge.

Am a bit rusty with command line github, but have followed instructions here https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-using-the-command-line I can't actually see any merge conflicts with my local file. I think the simplest approach will be to close this pull request and I'll make another one from a fresh branch?

OriolAbril:
> Am a bit rusty with command line github, but have followed instructions here https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-using-the-command-line I can't actually see any merge conflicts with my local file. I think the simplest approach will be to close this pull request and I'll make another one from a fresh branch?

It' probably easier to fix this way yes, notebooks and git diffs are not a good combination

PR #161, opened 2021-05-21
Heteroskedastic GPs notebook

JohnGoertz:
## Description
Several other GP frameworks (namely [GPflow](https://gpflow.readthedocs.io/en/master/notebooks/advanced/heteroskedastic.html) and [GPy](https://nbviewer.jupyter.org/github/SheffieldML/notebook/blob/master/GPy/heteroscedastic_regression.ipynb) have examples 
of how to model heteroskedasticity with GPs, but PyMC3 doesn't. There are hints scattered throughout the Discourse on how to do so via latent GPs through sparse approximations and/or coregionalization, but these largely rely on external links and aren't collected in one place.

I've put together a notebook that works through various ways to achieve this. Let me know if there are any changes or extensions you'd like to see!


twiecki:
I just skimmed but this looks awesome - thanks so much!

@drbenvincent or @bwengals: do you have time to take a look?

drbenvincent:
I can take a look. I've not used GP's, so my comments would probably be restricted to clarity / accessibility.

bwengals:
this is REALLY cool

JohnGoertz:
Okay, I've re-run with a higher `target_accept`, which reduced but didn't eliminate divergences. However, upon attempting to demonstrate how to get the learned correlation coefficient out of the LMC, it seems it's not really working as expected. It appears to have learned a slight negative correlation (-0.1), albeit very wide 95% CI bounds (-0.9,0.9). I added a note of this, as well as a discussion of the sampler statistics (particularly `ess_tail`), but maybe it's best to omit the LMC from this notebook? Perhaps this isn't the best example of its use, and it may be more confusing than helpful.

drbenvincent:
All my points were addressed, this all looks great to me. But I'm not experienced with GP's, so I think this notebook just needs a final check over from someone with more GP knowledge?

twiecki:
Thanks @JohnGoertz and @drbenvincent!

JohnGoertz:
That's great, thanks for looking it over and glad to contribute!

Issue #126, opened 2021-04-05
censored data (survival)


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/survival_analysis/censored_data.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

drbenvincent:
I'm happy if you want to assign this one to me

OriolAbril:
Done, thanks!

drbenvincent:
**TODO:** Add a link to the "Bayesian regression with truncated or censored data" notebook (#162). At the moment it is not clear what the URL will be. @OriolAbril suggests...
> "We should probably find a way to use custom anchors for links instead of relying on the url. Therefore, if we reorgnize the examples or we update the base url (i.e. adding a /stable/ /3.11.2/ verision indicator)... we only need to regenerate the documentation instead of hunting down all urls"

PR #163, opened 2021-05-22
update to Censored Data Models notebook 

drbenvincent:
Here are a few updates to the Censored Data Models notebook #126 
- Updated to best practice.
- Corrected an error in that the data in the first figure was truncated instead of censored.
- Added additional example to make it clearer that you get biased estimates using an uncensored model on censored data.
- Updated the prior over `mu` in the models. Previous prior was causing systematic underestimates in `mu`

It would be good to put a link in to the notebook on truncated and censored regression #162, but that's not been merged so not sure what the final URL will be.


OriolAbril:
> It would be good to put a link in to the notebook on truncated and censored regression #162, but that's not been merged so not sure what the final URL will be.

Can you comment that on the issue? We should probably find a way to use custom anchors for links instead of relying on the url. Therefore, if we reorgnize the examples or we update the base url (i.e. adding a /stable/ /3.11.2/ verision indicator)... we only need to regenerate the documentation instead of hunting down all urls

OriolAbril (approved):

drbenvincent:
> Can you comment that on the issue?

Done

PR #166, opened 2021-05-30
bump nbqa and pyupgrade

MarcoGorelli:
## Description

Addresses issue #xxx and aims to advance it to

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 - repo: https://github.com/nbQA-dev/nbQA
-  rev: 0.6.0
+  rev: 0.10.0
   hooks:
     - id: nbqa-black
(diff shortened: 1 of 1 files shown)


OriolAbril:
This will generate git conflicts with a couple other PRs if merged as is. If someone has time it would be great to go over the changes here and comment those that affect notebooks that currently have an open PR there. Then merge this with only the non problematic changes. Maybe @chiral-carbon ? Maybe it's just me but I hate having git conflicts on notebooks

MarcoGorelli:
> Maybe it's just me but I hate having git conflicts on notebooks

yeah same, they're an extra level worse than "regular" merge conflicts 

Alternatively, could put off the pyupgrade bump for now and just upgrade nbqa as its handling of magics is now more stable and this risks causing less trouble to contributors

Issue #83, opened 2021-03-30
GLM model selection


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/generalized_linear_models/GLM-model-selection.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

chiral-carbon:
@OriolAbril can I work on this?

OriolAbril:
Great!

drbenvincent:
This could do with some simplification to focus on the core aspects of model comparison. Specific proposals:
- remove interactive aspects of this notebook. They only work when running locally, not when viewing on the website. And in fact it is pretty clunky when I run the ipywidgets locally. Seems like a distraction.
  - remove interactive the dataset viewing
  - remove the interactive posterior plot viewing and just show static posterior predictions
- remove `plot_annotated_trace` and just use `az.plot_trace`
- why are we sometimes using raw data, sometimes the standardised data?

OriolAbril:
Always in favour of simplifying notebooks to have a clearer and more to the point scope. My main concern (not remembering the content of either notebook right now so tread lightly) is that it might end up being too similar to https://www.pymc.io/projects/docs/en/stable/learn/core_notebooks/model_comparison.html

drbenvincent:
Good point. Could potentially be expanded to include model comparison by Savage Dickey method perhaps

drbenvincent:
I didn't actually realise there were these separate core notebooks.

I think it's very cool to have these core notebooks under "Learn". But I think these notebooks should additionally be visible along with the regular pymc example notebooks. Some people might go straight to the Examples and not know about the extra ones under Learn.

Can I propose that these core notebooks stay where they are, but are additionally available with the rest of the Example notebooks?

OriolAbril:
[...]

PR #173, opened 2021-06-09
Update GLM Model Selection to best practices

chiral-carbon:
Addresses issue #83  and aims to advance it to Best Practices

- [x] Use numpy generator
- [x] Update watermark
- [x] Use bambi


chiral-carbon:
I am facing an error in cell 16, would like some help with resolving it.
Also, the results from manual model (cell 11) and using bambi (cell 13) are quite different, so I want some help with understanding that as well.

OriolAbril (commented):
Also, even though we basically have one model in this notebook, we have 3 different representations of it which I think can be quite confusing. math+widgets use one set of variable names, manual pymc3 code another and glm/bambi model a 3rd. 

Ideally these should all be the same or extremely similar and unambiguous, so the notebook is easier to follow, but I am not sure how bambi chooses names or if it's possible to customize those somehow, so before choosing which names to use I wanted to get an idea of the possibilities. cc @martinacantaro 

The model also has a lot of divergences, even in linear/quadratic versions, which probably has an effect to waic and loo which also show warnings always, so while the results don't make sense (i.e. k5 is supposed to be the best model for quad data) there isn't really any reason why they should make any sense. I noticed that only the x data is standardized, maybe standardizing y too (plus maybe undoing the transformation for the plots) works better? Other ideas?

OriolAbril (approved):

PR #177, opened 2021-06-10
glm bambi usage (WIP)

almostmeenal:
addresses issue #91 

- [x] use bambi instead of glm module
- [x] change imports and other possible bad practices, ensure notebook isnt broken and practices like inference_data=true are followed

Note: I've just started hopefully, i'll finish it in a day or two

cc: @OriolAbril

PR #179, opened 2021-06-18
improvements to glm.ipynb

mjhajharia:
glm.ipynb is not broken anymore after #177, a few minor updates in documentation or markdown cells are still left, doing them in this PR as i get time, nothing urgent, but everyone feel free to make suggestions!!!!


almostmeenal:
@OriolAbril I've taken care of all the comments, copying this comment from the previous PR over here so it stays in my head.

_

> The explanation is great and the link to the family object too. I also have a myst related extra note here (again, so not implement yet) that we could use intersphinx for that link so that if bambi decides to rename master to main, we'll only have to change a link in conf.py and regenerate docs for everything to work again, otherwise we'll have to track down all links and update them. Same thing if documentation is reorganized, files renamed or moved to a different folder structure

_

MarcoGorelli (approved):
looks good to me

fonnesbeck:
@almostmeenal just address Oriol's comment and this is good to merge.

almostmeenal:
thanks a ton for the help @OriolAbril !!!!

almostmeenal:
@fonnesbeck this is done, i'm sorry for the delay :((

PR #182, opened 2021-06-19
Simpson paradox example notebook

drbenvincent:
Hi. Here is a new notebook. There are already linear regression and hierarchical regression notebooks, but I think this one adds: a) a focus on Simpson's paradox (which is a popular and widely searched term), b) does a reasonable job of showing best practice, eg in `coords`, `xarray`, posterior predictions vs posterior predictive _distributions_, etc.

I'm very happy to get any feedback on the points below, or any other points in fact:
1. Am I right in thinking Model 1, 2, and 3 are presented by this notation? 1) `y ~ 1 + x`, 2) `y ~ 0 + (1 + x | group)`, 3) `y ~ 1 + x + (1 + x | group)` I'm not 100% convinced by the last one.
2. Referring to the 3-panel plots, what is the best language to use here? Posterior predictive vs posterior predictive _distribution_?
3. Should I add a more robust case for why the hierarchical model (model 3) is worth doing here or just refer to other sources?
4. Do you think it lacks a satisfactory conclusion, or is it fine as it is?
5. If anyone knows how to satisfy the `arviz` warnings then I'd be happy to hear. 
6. This is currently in `examples > GLM`, but let me know if you think it's better placed in `examples > Case Studies`
7. I believe I've used `coords` and `xarray` properly, but these are both still a little new to me, so I'm open to any simplifications/corrections.

### DIFF
--- a/examples/table_of_contents_examples.js
+++ b/examples/table_of_contents_examples.js
@@ -26,6 +26,7 @@ Gallery.contents = {
     "generalized_linear_models/GLM-robust": "(Generalized) Linear and Hierarchical Linear Models",
     "generalized_linear_models/GLM-rolling-regression": "(Generalized) Linear and Hierarchical Linear Models",
     "generalized_linear_models/GLM-truncated-censored-regression": "(Generalized) Linear and Hierarchical Linear Models",
+    "generalized_linear_models/GLM-simpsons-paradox": "(Generalized) Linear and Hierarchical Linear Models",
     "gaussian_processes/GP-Kron": "Gaussian Processes",
     "gaussian_processes/GP-Latent": "Gaussian Processes",
     "gaussian_processes/GP-Marginal": "Gaussian Processes",


MarcoGorelli:
Really nice!

fonnesbeck:
@drbenvincent looks really nice. Please just fix the ArviZ typo, and we can merge!

martinacantaro:
Task linked: [CU-5t5xv0 Read the Docs integration](https://app.clickup.com/t/5t5xv0)

drbenvincent:
Thanks! Hoping to make changes at some point this week, when I get a chance

drbenvincent:
Ok. Have fixed the typo and have also added a closing summary section. Let me know if that needs any edits...

MarcoGorelli (changes requested):
Thanks for updating!

TBH I'm really not keen on having

```import warnings
warnings.filterwarnings("ignore")
```
in the docs, it encourages people to put that in their code and then they miss deprecation warnings and their pipelines break when packages make new releases

If this is to silence the ArviZ warning, does it not work to use
```
pp_y[np.newaxis, :]
```
instead of
```
pp_y
```
?

drbenvincent:
It works for model 1 passing in `az.plot_hdi(xi, pp_y[np.newaxis, :], ...` as you suggest.

For Models 2 and 3, it works passing in `pp_y[:, g == i][np.newaxis, :]` 

Happy to do that, but it's as clear as mud so was leaning towards not doing it. Will commit those changes just now

MarcoGorelli:
>  it's as clear as mud

😆  true that...does anyone else following along have clearer suggestion?

MarcoGorelli (commented):
Found it, you can do
```
with linear_regression:
    pm.set_data({"x": xi})
    pp_y = pm.sample_posterior_predictive(trace, keep_size=True)["y"]
```
and then `pp_y` will also be in the format `(chain, draw, shape)`, and then `arviz` will take it without warnings and no `[np.newaxis, :]` will be necessary.

```
ax[1].plot(xi, np.mean(pp_y, axis=0), "k")
```
will still need changing though, e.g. to `axis=(0, 1)` (alternatively, you could add the posterior predictive to `trace` with `trace.add_groups(posterior_predictive={'observed': pp_y})` and then do `trace.posterior_predictive.mean(dims=('chain', 'draw')).to_array().squeeze()`, which is longer but at least avoids mysterious magic numbers)

drbenvincent:
Thanks for your help. Famous last words, but I think that's done it now!

I also filed an issue highlighting that PyMC should really return xarray objects for the predictive samples. https://github.com/pymc-devs/pymc3/issues/4863

MarcoGorelli (approved):
Looks good to me!



---

The only nitpick I have at this point is around the title "Random slopes and intercepts model", I just don't know how standard it is to call them like that - maybe "Independent slops and intercepts model"? Not a blocker though, at least not from my end

drbenvincent:
> The only nitpick I have at this point is around the title "Random slopes and intercepts model", I just don't know how standard it is to call them like that - maybe "Independent slops and intercepts model"? 

Agreed - I think it varies a lot by discipline. I've just gone with independent slopes and intercepts

Issue #43, opened 2021-03-15
PyMC3 Examples Known Errors & Notes (March 2021)


CloudChaoszero:
# Summary

The following are existing issues & suggestions in the pymc3-examples repo after going through an iteration of renaming plot dependencies from `pm.` to `arviz.`

> Note: This is similar to the pre-existing #34 issue.

## Installs
* Pymc3: v3.11
* theano-pymc (aesara):  v 1.1.2
# Issues

## General Issues

* Have param `returninference =True`

## File-specific Issues

- [ ] examples/pymc3_howto/lasso_block_update.ipynb

<details>
<summary>
Issue
</summary>

"`return_inferencedata` should be manually set, either to True or to False, but one of the two to avoid the warning.

If False, an `inferencedata` should be created within the model context, and passed to arviz. This will 
1. avoid the warning of conversion without the model context and 
2. push forward arviz best practices, it is probably not too relevant here but conversion may not be cheap for some models because it requires computing all the pointwise log likelihood values. `az.plot_xyz(trace)` works because ArviZ internally converts the data to inferencedata, then plots."
</details>

- [ ] examples/pymc3_howto/data_container.ipynb

<details>
<summary>
Issue
</summary>
We should have "keep_size=True to avoid the warning in the cell below, also because in a future ArviZ release the behaviour of hdi will change for 2d arrays (not for 1d or 3d+ arrays), so using 3d arrays with chain, draw, *shape should be used."
</details>

- [ ] examples/pymc3_howto/sampling_conjugate_step.ipynb

<details>
<summary>
Code
</summary>

```python
traces = []
models = []
names = ["Partial conjugate sampling", "Full NUTS"]

for use_conjugate in [True, False]:
    with pm.Model() as model:
        tau = pm.Exponential("tau", lam=1, testval=1.0)
        alpha = pm.Deterministic("alpha", tau * np.ones([N, J]))
        p = pm.Dirichlet("p", a=alpha)

        if use_conjugate:
            # If we use the conjugate sampling, we don't need to define the likelihood
            # as it's already taken into account in our custom step method
            step = [ConjugateStep(p.transformed, counts, tau)]

        else:
            x = pm.Multinomial("x", n=ncounts, p=p, observed=counts)
            step = []

        trace = pm.sample(step=step, chains=2, cores=1, return_inferencedata=True)
        traces.append(trace)

    assert all(az.summary(trace)["r_hat"] < 1.1)
    models.append(model)
```

</details>

[...]

Issue #12, opened 2020-12-22
GP-MeansAndCovs.ipynb throws LinAlgError


MarcoGorelli:
Cell 6 contains:
```python
lengthscale = 0.2
eta = 2.0
cov = eta ** 2 * pm.gp.cov.ExpQuad(1, lengthscale)

X = np.linspace(0, 2, 200)[:, None]
K = cov(X).eval()

plt.figure(figsize=(14, 4))

plt.plot(X, pm.MvNormal.dist(mu=np.zeros(K.shape[0]), cov=K).random(size=3).T)
plt.title("Samples from the GP prior")
plt.ylabel("y")
plt.xlabel("X");
```
which throws
```
ValueError: input operand has more dimensions than allowed by the axis remapping
```

As per #11 , I can fix this by specifying the shape of the MvNormal:
```
lengthscale = 0.2
eta = 2.0
cov = eta ** 2 * pm.gp.cov.ExpQuad(1, lengthscale)

X = np.linspace(0, 2, 200)[:, None]
K = cov(X).eval()

plt.figure(figsize=(14, 4))

plt.plot(X, pm.MvNormal.dist(mu=np.zeros(K.shape[0]), cov=K, shape=K.shape[0]).random(size=3).T)
plt.title("Samples from the GP prior")
plt.ylabel("y")
plt.xlabel("X");
```
but then I get
```python-traceback
~/pymc3-dev/pymc3/distributions/multivariate.py in random(self, point, size)
    277 
    278         if self._cov_type == "cov":
--> 279             chol = np.linalg.cholesky(param)
    280         elif self._cov_type == "chol":
    281             chol = param

<__array_function__ internals> in cholesky(*args, **kwargs)

~/miniconda3/envs/pymc3-dev-py38/lib/python3.8/site-packages/numpy/linalg/linalg.py in cholesky(a)
    762     t, result_t = _commonType(a)
    763     signature = 'D->D' if isComplexType(t) else 'd->d'
--> 764     r = gufunc(a, signature=signature, extobj=extobj)
    765     return wrap(r.astype(result_t, copy=False))
    766 

~/miniconda3/envs/pymc3-dev-py38/lib/python3.8/site-packages/numpy/linalg/linalg.py in _raise_linalgerror_nonposdef(err, flag)
     89 
     90 def _raise_linalgerror_nonposdef(err, flag):
---> 91     raise LinAlgError("Matrix is not positive definite")
     92 
[...]

PR #195, opened 2021-07-23
[WIP] Advance to Best Practices: Multivariate Gaussian Random Walk Notebook

ltoniazzi:
## Description

Addresses issue #140 and aims to advance it to Best Practices

## Best Practices
General updates: everything addressed.
ArviZ: 
- please doublecheck my usage of xarray.
- Not sure how to use coords/dims effectively in this notebook.

## Changes for discussion
- Emphasized regression modelling.
- Added docstrings and text between code blocks.
- Added two ArviZ plots.
- How to justify the `Scaler` (like with a reference about sampling improvement)?
- When I run `az.summary(trace).r_hat <1.03` one entry of a Cholesky matrix scores 1.04. Is that ok? (referring to this [guide](https://github.com/pymc-devs/pymc3/wiki/PyMC3-Jupyter-Notebook-Style-Guide))

Cheers

---
Comment: I'll be available to work on this again on the 22nd of August.


examples/time_series/MvGaussianRandomWalk_demo.ipynb
```diff
@@ -2,6 +2,7 @@
  "cells": [
```
MarcoGorelli, line 2: <pre class="ql-syntax" spellcheck="false">Line #2.    α_mean = trace.posterior["α"].mean(axis=(0, 1))
</pre><p>rather than <code>axis=(0, 1)</code></p><p>which is a bit mysterious, would <code>.mean(dim=('chain', 'sample'))</code> work?</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/195/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

ltoniazzi: Used `dim=("chain", "draw")`.

OriolAbril, line 2: <p>note to self (or to however has time to check): I am not sure sphinx/myst support the <code>\begin{}</code> latex syntax out of the box, we may need some extra tweaks/configuration. I vaguely recall reading something about that, so I am quite sure it won't be a problem once configured and that how to configure that will be somewhere in https://myst-parser.readthedocs.io/en/latest/, maybe in optional syntaxes?</p><p> </p><p>myst should also support adding labels to equations IIRC which I think will be more convenient than the manual star sign.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/195/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

OriolAbril, line 2: <p>the link to api docs should ideally be:</p><p> </p><pre class="ql-syntax" spellcheck="false">{class}link for the documentation &lt;pymc3.distributions.multivariate.LKJCholeskyCov&gt;
</pre><p> </p><p>by doing this, if we restructure the api page or get multiversion docs both of which will probably change the exact url, we only need to rebuild the documentation for all the links to be fixed instead of having to track down all the urls throughout the notebooks.</p><p> </p><p>The link to the other notebook should probably be a <code>{doc}</code>  or <code>{ref}</code>  object, but I'm not yet sure which approach to recommend.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/195/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

OriolAbril, line 2: <pre class="ql-syntax" spellcheck="false">Line #43.            ppc = pm.sample_posterior_predictive(trace=trace)
</pre><p> </p><p>I would directly add this to the inferencedata. I think the following should work.</p><p> </p><pre class="ql-syntax" spellcheck="false">trace.extend(az.from_pymc3(posterior_predictive=pm.sample_posterior_predictive(trace)))
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/195/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

OriolAbril, line 2: <p>I would separate sampling and plots in different cells and split the explanation above so they are closer to the relevant plot. </p><p> </p><p>Maybe even add extra links or references, i.e. to the posterior predictive notebook when using plot_ppc. This probably depends on what is assumed as known to readers of this notebook.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/195/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

OriolAbril, line 2: <pre class="ql-syntax" spellcheck="false">Line #12.    plt.gca().set_prop_cycle("color", ["red", "green", "blue"])
</pre><p> </p><p>the colors used should be the ones of the theme, or alternatively, use <code>C1</code>,  <code>C3</code>,  <code>C0</code>  in plots to alter the order. red and green look the same to some colorblind people.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/195/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

OriolAbril, line 2: <p>I'm not sure if the scaler will be compatible with the ppc being stored as an xarray dataset, but when doing that, and using meaningful dimension names, the <code>[:, :, 0]</code>  becomes <code>.sel(dim_name="first_label")</code>  or <code>isel(dim_name=0)</code>.  I don't know which names to use though.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/195/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

OriolAbril (commented):
Thanks for the PR, this makes the notebook much more clear

examples/time_series/MvGaussianRandomWalk_demo.ipynb
ltoniazzi: <p>Sorry, I do not understand how to implement the link with your suggested method. Can you link a notebook with an example of this approach?</p>

ltoniazzi: <p>Not sure how to use colors from the theme so that I can match <code>data</code> and <code>pred</code> colors. Also, if I set more than one color in a plot, for examples with <code>plt.plot(t, y, ".", color = ("C0", "C1", "C3"))</code>, I get the error <code>Cannot put cycle reference ('C0') in prop_cycler</code>.</p>

OriolAbril: We recently changed the notebook renderer from nbsphinx to myst-nb which is built on top of the myst parser which is extended markdown to basically support all of rST features. We can therefore use any sphinx directive and role from the notebooks. I added some guidance on sphinx hyperlinks using either markdown or rst at 

As for examples, https://arviz-devs.github.io/arviz/user_guide/pymc3_refitting_xr_lik.html has at least one of these intersphinx links and https://github.com/pymc-devs/pymc3/blob/init_glossary/docs/source/developer_guide_implementing_distribution.md is pure markdown but has multiple examples and there are no rendering differences between the markdown in notebooks and markdown files anymore, everything used there can also be used in a markdown cell of a notebook

OriolAbril: I see. I think the best way to go is calling plot on `data` directly, then use `plt.gca().set_prop_cycle(None)` to reset the color cycle and then call plot on `pred`. This will ensure both plots use c0, c1 and c2 which are the first 3 colors of the theme and a colorblind friendly color pallette. 

Extra nitpick, given that we use the axes object however, I personally would prefer using `ax.` instead of the `plt` interface.

ltoniazzi: I did this and introduced `coords`. Everything seems to work but I get an odd message after the ppc sampling, namely the output of `In [6]` of the last commit says:
```
0, dim: steps, 300 =? 300
1, dim: y_, 3 =? 3
```

OriolAbril: This is a forgotten debug print that made it into latest arviz release, it has already been fixed in main but the new release is still pending

OriolAbril: This has an example of a sphinx label used on one equation: https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#dollar-delimited-math

And it is indeed possible to use `\begin` syntax directly (see https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#syntax-amsmath), but I'm not sure if labels will work. I think for now the best would be to put the latex code inside a double dollar env or a math directive as shown in the first example

ltoniazzi: Sorry, I do not know how to compile these commands to check if they work correctly, referring also your previous comment on  sphinx labels (I never used sphinx). Can you suggest a source for this?

OriolAbril: <p>Resources to learn sphinx are historically bad and impossible to find. I know there are a lot of people working on that so there should be great tutorials coming up soon, but I think they are still WIP. In my opinion however, it's much more important to learn myst which is what we'll basically use here and/or rST (used in docstrings) than it is to actually learn sphinx itself. The docs here are already set and configure to be built with more or less everything we want so changes to sphinx configuration or it's extensions are quite rare.</p><p> </p><p>I like https://myst-parser.readthedocs.io/en/latest/ but I was already relatively familiar with rST so I might be a bit biased. Let me know what you think about these docs. </p><p> </p><p>Regarding this specific case, which is also quite central and common when writing docs, what I am suggesting is to update the "raw" link that currently looks like:</p><p> </p><pre class="ql-syntax" spellcheck="false">[link for the documentation](https://docs.pymc.io/api/distributions/multivariate.html#pymc3.distributions.multivariate.LKJCholeskyCov)
</pre><p> </p><p>for a cross-reference (very similar to a latex <code>\ref</code>  for example). As this is a cross reference, it won't be rendered by the markdown preview, again, like latex, it will only be rendered after building the whole documentation with sphinx (we have some docs at https://github.com/pymc-devs/pymc3/wiki/Generate-Documentation but hopefully all that will soon be better documented and overall easier). </p><p> </p><p>For now, I would live dangerously and trust my gut that things will eventually render correctly. </p><p> </p><p>In myst, cross-references use the following patterns:</p><p> </p><pre class="ql-syntax" spellcheck="false">{reference_type}<code>reference_target</code>
# or
{reference_type}<code>custom text &lt;reference_target&gt;</code>
</pre><p> </p><p>here I think we should use the 2nd option with the custom text. Our target is the LKJ class and as we are defining the target as the import path we have to set the reference type to class.</p>

ltoniazzi: Thanks, extremely clear explanations! I gave it a go adding reference_types: func, class and doc. 

Remarks:
1. for [this notebook reference](https://docs.pymc.io/pymc-examples/examples/diagnostics_and_criticism/posterior_predictive.html) I tried some nonsense with 
```
{doc}`posterior predictive samples <pymc-examples.examples.diagnostics_and_criticism.posterior_predictive>`
```
2.  I could not guess the reference_target of [this notebook](https://docs.pymc.io/notebooks/LKJ.html).

OriolAbril, line 2: <p>doc type references should use paths, and they can use paths from the home directory of the documentation. As here we have the <code>conf.py</code>  within the examples directory, this is set as the "home" of the docs. You can always double check the references with:</p><p> </p><pre class="ql-syntax" spellcheck="false">sphobjinv suggest -t 90 -u https://pymc-examples.readthedocs.io/en/latest/ posterior_predictive
&nbsp;
No inventory at provided URL.
Attempting "https://pymc-examples.readthedocs.io/en/latest/objects.inv" ...
Remote inventory found.
&nbsp;
:std:doc:<code>diagnostics_and_criticism/posterior_predictive</code>
</pre><p> </p><p>so here, using </p><p> </p><pre class="ql-syntax" spellcheck="false">{doc}<code>diagnostics_and_criticism/posterior_predictive</code>
</pre><p> </p><p>should work</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/195/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

OriolAbril:
@ltoniazzi can you rebase on main to get rtd to generate a preview of the notebook? I am quite sure the references will work except the doc one but it would be great to check they really do

examples/time_series/MvGaussianRandomWalk_demo.ipynb
OriolAbril: I answered the 1st point below before seeing this.

Regarding the 2nd point, `doc` type references use paths and those can either be relative to the home of the docs or to the current file. I find relative to the home of the docs much better, so it would be `` {doc}`case_studies/LKJ` ``

reference https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#cross-referencing-using-roles

OriolAbril:
Looks great so far and shows how myst "unlocking" sphinx references on notebook markdown is extremely useful 😄

examples/time_series/MvGaussianRandomWalk_demo.ipynb
ltoniazzi: Awesome thanks!  Just noting that `{doc}`case_studies/LKJ` did not catch the home of the docs folder, while `{doc}`/case_studies/LKJ` did (also for the other doc reference in the notebook).

OriolAbril (approved):

Issue #82, opened 2021-03-30
GLM logistic


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/generalized_linear_models/GLM-logistic.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

twiecki:
This is also using the deprecated `glm` submodule which we should remove and implement a logistic regression directly.

chiral-carbon:
hey! can I take this up?

twiecki:
Totally, thanks for offering to help.

drbenvincent:
I can update this to v4

drbenvincent:
@OriolAbril / @twiecki 
I've started to port this, but noticed that it overlaps _a lot_ with an official Bambi example (see https://bambinos.github.io/bambi/main/notebooks/model_comparison.html)

I will definitely link to that, but I'm wondering if this notebook is essentially redundant?

Let me know if we should kill this or just continue to port what is there.

cfonnesbeck:
I vote in favor of removing. We should take advantage of any opportunity to reduce the size of the example library.

aloctavodia:
Agreed. Is there a place in PyMC were we redirect/point to bambi for GLMs?

drbenvincent:
Ok. Will submit PR to remove this

> Agreed. Is there a place in PyMC were we redirect/point to bambi for GLMs?

Some of the GLM notebooks do hand coded PyMC models. Some include Bambi only. Some contain a mix.

If we were going to move towards being consistent, my preference would be to focus on hand coded PyMC models but with linking to Bambi and possibly including non-executed Bambi code in markdown cells.

[...]

PR #196, opened 2021-07-26
[WIP] Update glm-logistic with bambi

chiral-carbon:
Addresses issue #82 and aims to advance it to best practices, use `bambi` instead of pymc glm module.

I had a question regarding the family to be used in cells 4 and 14 while defining the model. The notebook originally uses Binomial family with the GLM module but with `bambi` I think we do not have a default binomial family defined. I used `negativebinomial` just in case but since it is not the correct replacement, what should I do about it?


tomicapretto:
Hi @OriolAbril, thanks for tagging me here! 

Actually, I realized it wasn't so much work to implement some stuff to have the Binomial family. It's still being implemented in formulae (see https://github.com/bambinos/formulae/pull/40), and it will be implemented in Bambi after that. 
As for the timeline, this should be done for the next release which will happen when GSoC ends.

chiral-carbon:
thanks a lot @tomicapretto , means a lot to me! :smile:

tomicapretto:
Hi @chiral-carbon. Bambi development version now has a Binomial family :tada: 
Have look [here](https://tcapretto.netlify.app/post/2021-08-03-binomial-family-in-bambi/) to see how it works (and how it compares to the existing Bernoulli family)

chiral-carbon:
this is great @tomicapretto!

OriolAbril (approved):

PR #197, opened 2021-07-30
update glm rolling regression

chiral-carbon:
Addresses issue #89 and aims to use `bambi` instead of pymc3 GLM module.


twiecki:
I feel like we should just do the linear regression by hand here instead of using bambi, that'd be more obvious.

OriolAbril:
There is also an issue with many broken links throughout the notebooks linked in https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview

chiral-carbon:
not sure if this needs a `references` section?  the link to pairs trading in the beginning of the notebook is the only external reference, should I include it in `references.bib` file?

OriolAbril (approved):

PR #202, opened 2021-08-02
Update GLM Linear to use bambi

chiral-carbon:
Addresses issue #81 and aims to replace usage of `pymc3` GLM module with `bambi`


tomicapretto:
@OriolAbril @chiral-carbon. Bambi's current release does fit a model with statsmodels and uses the estimations from that fit to propose default prior distributions. **However**, that is not going to be the default behavior for the release that is about to come. In this new release, we use an approach very similar to the one in [rstanarm](https://mc-stan.org/rstanarm/articles/priors.html)

Finally, @chiral-carbon, please don't hesitate to ping me if you need help with something related to Bambi or you want to double-check something. If you're in either ArviZ or PyMC3 Slack feel free to contact me there  :smile:

twiecki (approved):

MarcoGorelli (approved):

Issue #53, opened 2021-03-29
LKJ


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/case_studies/LKJ.ipynb
Reviewers: @AlexAndorra

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### ArviZ related
* Use labeled dims in code cells 12 and 13. Also use `.values` instead of `.data`.

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### ArviZ related
* Show how to index correlation matrixes with xarray. See https://github.com/arviz-devs/arviz/issues/1627 and https://arviz-devs.github.io/arviz/user_guide/label_guide.html#custom-labellers for details and examples on this (note that ArviZ docs are for the development version, at the time of writing, labellers are an unreleased feature)

### Exotic dependencies
None

### Computing power
All models run in less than a minute

chiral-carbon:
taking up this issue.

PrinceAsiedu:
Working on this one

twiecki:
Looks done.

PR #206, opened 2021-08-09
Update case study: LKJ

chiral-carbon:
Addresses issue #53 and aims to advance it to best practices.

I did not see much to update here apart from using `numpy.random.default_rng()` so let me know if any improvements could be made in the code or plotting.


OriolAbril:
> I did not see much to update here apart from using numpy.random.default_rng() so let me know if any improvements could be made in the code or plotting.

also fix the comment on the issue description about using `.values` instead of `.data`

MarcoGorelli (approved):
looks good to me (@OriolAbril not sure how actionable the comment about all the warnings is at the moment or whether that's OK to leave as a follow-up, will leave this with you 😄 )

OriolAbril:
We'll go over the two minor nits still missing in my opinion and merge in the next few days. Trying to make the final plot a bit more clear and see if we can remove some (or even all) warnings in the plot_trace figure

OriolAbril (approved):

Issue #102, opened 2021-04-05
gaussian mixture model


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/mixture_models/gaussian_mixture_model.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

drbenvincent:
This could be updated to use new `pm.Mixture`

drbenvincent:
Any reason why the cluster means can't use `pymc.distributions.transforms.Ordered`?

drbenvincent:
Actually, anyone mind if I update this with `pm.NormalMixture`? We can get some significant simplification of this notebook.

OriolAbril:
That sounds amazing and things like that are the reason there is a `v4 (auto)` and a `Done` category in the [tracker project](https://github.com/pymc-devs/pymc-examples/projects/1) that are clearly separated. It wasn't the case, but this notebook might have run when changing `import pymc3` to `import pymc` and theano to aesara, but that a notebook runs on v4 doesn't mean it is updated to be _documentation_ for v4.

I am not sure how much of that was already available in latest v3 releases, but the issue is the same. A notebook is added to cover and explain a specific use case, a while later we see (or have time to implement) a way to make that case better and easier, but when the feature is added to pymc, the notebook is not updated. When that happens many users won't be aware the new feature exists, and others might now but be unable to use it without an example notebook so the new feature ends up being underused and the old but well documented way of doing things prevails.

drbenvincent:
Great. Will put in a PR soon.

drbenvincent:
Moving this in the notebook state tracker from 'todo' to 'done'. And closing issue

PR #208, opened 2021-08-11
update gaussian mixture model notebooks

chiral-carbon:
Addresses the issues: #102 and #103 to advance them to best practices.

### DIFF
--- a/examples/mixture_models/gaussian_mixture_model.ipynb
+++ b/examples/mixture_models/gaussian_mixture_model.ipynb
@@ -7,8 +7,8 @@
     "# Gaussian Mixture Model\n",
     "\n",
     "\n",
-    ":::{post} Sept 17, 2021\n",
-    ":tags: mixture model\n",
+    ":::{post} Sept 18, 2021\n",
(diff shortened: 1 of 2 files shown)


OriolAbril:
1. Count how many occurences of each integer are along all chains and draws for each observation. This will generate a 2d array with dims (obs_id, mixture_component)
2. Sort along the mixture component and keep only the second column
3. Find the argmax. This is the observation whose the 2nd mixture component has a higher weight so it should be the less clear one

OriolAbril:
> The values between 0 and 500 are indices for data. So, as Oriol said, we need to find data points close to -2 and 2. On top of that, we need their indices so that they can be passed in coords={"obs_id": [...]} .
>
> Rephrasing -
>
> data close to -2
> data + 2 close to 0
> |data + 2| close to 0  # Taking absolute to make values on R+ scale.
> argmin(|data + 2|)  # Will give the index
>
> So, I think something like np.argmin(np.abs(data + 2)) will return the index for data point close to -2. And the same can be done for any data point of interest.

I was proposing a different approach though so that we don't rely on the initial exploration that -2/2 look like boundaries and instead take the results and plot the observation to which the model assigned more "weight" to the 2nd category. Both will probably be fine so feel free to try whichever looks easier and if it were to not work try the other.

chiral-carbon:
@OriolAbril  I had tried @Sayam753's method and have used that to get points that have greater competition between clusters. do they look okay?

OriolAbril (approved):

Issue #101, opened 2021-04-05
dirichlet process mixtures for density estimation


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/mixture_models/dp_mix.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

OriolAbril:
I tried moving from nbsphinx to myst-nb and was unable to do so because the widgets in this notebook are not saved correctly, see https://github.com/jupyter/jupyter-sphinx/blob/master/jupyter_sphinx/ast.py#L607. After manually editing the metadata to remove the widgets completely I was able to build the documentation.

Side note: the DE-MCMC notebooks _also_ have widgets but they are correctly saved are are not a problem.

chiral-carbon:
can I work on this, since the previous PR only made metadata changes?

OriolAbril:
Yes,  it still needs to be updated. I only fixed the metadata because it was broken and was blocking the switch from nbsphinx to myst-nb

LegrandNico:
Hi, I would like to update this notebook to pymc v4, is there anything I should know?

PR #212, opened 2021-08-18
Update dirichlet mixture model notebooks

chiral-carbon:
Addresses issues #100 and #101 and aims to advance the notebooks to best practices.


examples/mixture_models/dirichlet_mixture_of_multinomials.ipynb
```diff
@@ -65,8 +65,9 @@
    },
```
chiral-carbon, line 65: <p>tried running the notebook first without making any major changes (except using numpy random Generator for setting the seed), but I get an error here. The <code>summary_multinomial</code>  dataframe that I get is not the same as before, because,</p><p>creation of the column <code>ess_mean_per_sec</code> depends on a column called <code>ess_mean</code>. However, here we don't get the columns <code>ess_mean</code>  and <code>ess_sd</code> , is there something obvious I missed here?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/212/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

OriolAbril: <p>arviz summary now uses <code>ess_bulk</code>  as default. They are not exactly the same but nearly. For the purposes of these notebook we can just change <code>ess_mean</code>  for <code>ess_bulk</code>  </p>

OriolAbril, line 65: <p>Use coords. This example tries to be more down-to-earth even if using simulated data so the number of clusters (k again) corresponds to the tree species, so it would be great to use as coord values real tree names :). For the number of observations, here they now correspond to the number of forests, so again, as there are only 10 "forests" we could also use real forest names.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/212/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

OriolAbril, line 65: <pre class="ql-syntax" spellcheck="false">Line #3.            draws=int(5e3), chains=4, step=pm.Metropolis(), return_inferencedata=True
</pre><p><code>5000</code>  is shorter than <code>int(5e3)</code>  and I find it more clear</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/212/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

chiral-carbon: <p>ah got it, that's a cool idea. will add that!</p><p> </p>

OriolAbril (approved):

Issue #184, opened 2021-06-20
Guide for writing custom distributions


ally-lee:
I would like to contribute a guide for defining custom PyMC3 distributions. I was working with count data and realized that the Poisson distribution didn't offer the flexibility in dispersion I needed since the mean and variance are fixed to each other. I decided that the Generalized Poisson was a better likelihood for my model, so I wrote a custom distribution for it. I have put together a notebook outlining this, and I'll link my PR to this issue.

I posted about this on discourse a while back, so it would be great to post updates for the community in this thread:
https://discourse.pymc.io/t/generalized-poisson-distribution/6535

File:
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

drbenvincent:
Just flagging up that this notebook is still needs to be updated from PyMC3

PR #220, opened 2021-09-01
Update pymc3_howto/custom_distribution NB to best practices

chiral-carbon:
Aims to advance https://github.com/pymc-devs/pymc-examples/issues/184 to best practices.


OriolAbril:
> it renders without the dollar sign though in jupyter when I see it locally. will remove the expression from being inline and push again

jupyter, reviewnb and sphinx (via myst) are 3 different ways of rendering notebooks. For our purposes here, what matters is sphinx rendering to work properly. Reviewnb and jupyter should be the same rendering, but sphinx is by design a completely different paradigm as it doesn't render notebooks independently but as a whole and it renders them into a website using extensions and a theme.

OriolAbril:
the references don't look right, you should double check the bibtex source (you can use google scholar to generate it for you). Right now some have 3 somewhat repeated links but not all of them work

chiral-carbon:
I used the bibtex citations provided by the website of the paper themselves, using the doi links does open the correct website so I'm not sure what to change

fonnesbeck:
@chiral-carbon what is the status of this? Is it close to done, or should I close it?

chiral-carbon:
working on it, will update here in a few days

drbenvincent:
Can you remove all `pymc.*` or `pymc3.*` tags in the first code cell?

PR #224, opened 2021-09-09
replace nbqa-black with black-jupyter

MarcoGorelli:
## Description

Replaces `nbqa`'s hook with black's built-in support for Jupyter, as it'll be far better maintained 

None of the modified notebooks are being modified by open PRs, should I think this should be OK to merge without it being disruptive


OriolAbril (approved):
looks great

PR #229, opened 2021-09-25
Binning example notebook

drbenvincent:
This pull request is for a novel example notebook. It covers how to make inferences about the parameters of a parametric distribution from observations of that distributions which have been binned. The notebook covers a range of situations, including making inferences about population parameters based upon two studies with binned data, but that have been binned _differently_. A collaborative effort from @ericmjl and myself.


drbenvincent:
Anything else needed on this one before we can merge?

OriolAbril:
can you update it to follow the latest additions tot the style guide? https://pymc--5224.org.readthedocs.build/en/5224/contributing/jupyter_style.html (it is already available in latest too, but there is has some rendering issues)

I think the relevant sections are first cell and authorship ones. I would also recommend using more "conceptual" tags in addition to case study. Maybe "binned data", something about trying to make inference on populations? It might also be interesting for readers to have a pymc labs tag so notebooks about the work there are easy to find, and being an institutional partner I think it's perfectly fine.

OriolAbril (approved):

OriolAbril:
Notebook is up at https://docs.pymc.io/projects/examples/en/latest/case_studies/binning.html

PR #231, opened 2021-09-29
Standardize the book recommendations section

olgadk7:
1. Got rid of the "PyMC3 port of the book" in the book titles
2. Unlink the author in three bullets
3. Add author in one bullet 
4. Add description in one bullet 

To do: switch PyMC3 to PyMC, including in logo


OriolAbril (approved):
Thanks. 

I copied that from https://github.com/pymc-devs/pymc at some point IIRC, so it might be worth checking and if needed open a PR there too. Also in general open to suggestions for this readme if you think it should be further reduced for exampe

PR #233, opened 2021-10-03
Update PULL_REQUEST_TEMPLATE.md

canyon289:
## Description

Addresses issue #xxx and aims to advance it to

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,16 +1,21 @@
 ## Description
-<!-- Thank you so much for your PR!
+<!-- Thank you so much for your PR to pymc-examples!
 
-To help us review your contribution, please provide the info requested below. Otherwise,
-we may leave reviews that are actually out of the scope of the current PR. -->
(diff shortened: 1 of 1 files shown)


OriolAbril (commented):
Can we keep a checklist item pointing to the relevant tracker issue? It has been a great help to have everything in https://github.com/pymc-devs/pymc-examples/projects/1 and I think it will still be extremely helpful with the rerunning+updating to pymc 4.0 so I think we should keep if for now and continue to open tracker issues for new notebooks

canyon289:
Updated link, In regards to tracker we can add it to checklist but can non maintainers edit that tracker? Or should the item say "ask a maintainer to add a card before merge?"

OriolAbril:
I want a checklist iitem that says "related to issue #xxx". There should be no PRs with no link to an issue (either tracker one or proposal one) unless they are part of the team. Then us maintainers have to go on the issue if a tracker one and update its position in the project (I don't know if it's possible to do automatically but if it is I don't know how) but it's much easier to do if the issue is linked. And doing this also leaves the "update history" of the notebook in the tracker, i.e. https://github.com/pymc-devs/pymc-examples/issues/81

Issue #234, opened 2021-10-03
Splines in PyMC


OriolAbril:
**Notebook url**: https://github.com/pymc-devs/pymc-examples/blob/main/examples/splines/spline.ipynb
reviewer: jhrcook

OriolAbril:
The notebook needs to be updated to follow the [style guide](https://github.com/pymc-devs/pymc/wiki/PyMC-Jupyter-Notebook-Style-Guide). All notebooks **must** have a post directive in their first cell.

Ideally it would also use matplotlib directly instead of plotnine to reduce the dependencies on external libraries

michaelosthege:
The notebook should also mention the `SplineWrapper` and explain the difference to the spline built in the notebook.
https://github.com/pymc-devs/pymc/blob/073e26b9c650ff9ae851ab9904cb3ad9c905a643/pymc/distributions/dist_math.py#L316-L319

OriolAbril:
Updated by https://github.com/pymc-devs/pymc-examples/pull/274, moving it to v4 auto to signal there are still changes needed on the metadata. 

See https://github.com/pymc-devs/pymc-examples/pull/274#discussion_r867411416

kuvychko:
Working on this notebook, as part of PyMC Open Source session #2 with @PrinceAsiedu

PR #232, opened 2021-10-03
Spline example

jhrcook:
## Description

Addresses issue #230 and aims to advance it to Best Practices.

This PR is for a contribution of an example of a spline regression in PyMC3.

Issue #198, opened 2021-07-31
Standardize authorship attribution


OriolAbril:
Right now it's not clear how authors of examples should claim authorship so they can get proper credit. There are several alternatives around:
* Table with version history: https://oriolabril.github.io/pymc3/pymc-examples/examples/generalized_linear_models/GLM-robust-with-outlier-detection.html?highlight=author#notes
* Authors on top of the blogpost: https://oriolabril.github.io/pymc3/pymc-examples/examples/getting_started.html?highlight=author, https://oriolabril.github.io/pymc3/pymc-examples/examples/case_studies/mediation_analysis.html?highlight=author. This would probably be compatible with ablog authorship. But I (Oriol) wasn't planning on exposing the page grouping examples by author and don't really see the use for it in this particular case.
* Authorship description at the end of the example, listing also authors of major updates, but not those of minor updates/rerunning of notebooks. https://oriolabril.github.io/pymc3/pymc-examples/examples/survival_analysis/bayes_param_survival_pymc3.html?highlight=author#authors or https://oriolabril.github.io/pymc3/pymc-examples/examples/case_studies/blackbox_external_likelihood.html?highlight=author#authors

We could even use https://github.com/sloria/sphinx-issues with the user, pr  roles to simplify linking to github users and relevant PRs if using approaches 1 or 3.

OriolAbril:
I propose to use option 3 with bullet points, like in https://pymc-examples.readthedocs.io/en/latest/survival_analysis/bayes_param_survival_pymc3.html. I think the dates are not necessary but don't hurt either. I propose to use the following categories:

* Alice **wrote**: for notebooks created specifically for pymc-examples
* Alice **adapted** Bob's blogpost: for notebooks adapted from blogposts. both authors and blogpost should be links to relevant places
* Alice **updated** the notebook: basically for everything
* Alice **extended** the notebook: (we could do without this one) for cases where someone adds new content to the notebook instead of making changes to improve the notebook without changing it's original intent. Updating could be a typo but could also be much more work than extending.
* Alice **re-executed* the notebook with PyMC3 x.y.z for rerunning the notebook without changes to code or text.

Main open question that comes to mind is if going forward we want those bullet points to link to the PR with the changes

PR #238, opened 2021-10-08
Convert plots in Spline example to matplotlib

tjburch:
I noticed in the (very well done!) spline example from @jhrcook added in #232 that plots were not in Matplotlib, but following the PR it seemed the preference was for matplotlib for ease of maintenance over time. I went ahead and converted the figures to matplotlib. 

Aside from the matplotlib changes, the lowess and knot-based lm portions had to be implemented in `statsmodels` before plotting, since those aren't available directly in mpl.

Tracking issue: #234

### DIFF
--- a/examples/splines/spline.ipynb
+++ b/examples/splines/spline.ipynb
@@ -1474,6 +1474,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Authors\n",
+    "\n",
     "- Authored by Joshua Cook in October, 2021\n",
     "- Updated by [Tyler James Burch](https://github.com/tjburch) in October, 2021"
    ]


tjburch:
Added in latest commit

OriolAbril:
Thanks! I have commented a few nits, nothing too significant

tjburch:
>It is also great that you have added yourself to the notebook, like @jhrcook did. As not everybody does, we were thinking about adding that the the style guide and setting a common style for it, both of you for example have not linked to your github profile now website like other authors do. If you can it would be great to have your opinions and preferences at #198 which is where we are discussing this.

Made a comment over on #198 that you've already seen. I went ahead and adapted the authorship comments here to reflect the 3rd option mentioned on that issue.

Issue #200, opened 2021-07-31
Guidance/standardization of "front matter"


OriolAbril:
I remember at some point discussing the possibility to in addition to the tags and categories that will already give an idea of the notebook topics and difficulty, we could define some kind of standard front matter so that all notebook have a quick explanation about what knowledge is assumed, links to notebooks that explain those concepts that are treated as known by the reader..

Sayam753:
Cross posting from slack -
We need to also standardise on how to handle external dependencies for notebooks. A simple idea is to do a `pip/conda install` at the start of notebooks.

OriolAbril:
"frontmatter" proposal:

```
(notebook_id)=
# Notebook title

|badges| |binder| |colab| (hopefully automated)

:::{post} date
:tags: tag1, tag2
:category: level
:::

:::{dropdown} Notebook dependencies
:icon: package-dependencies

In order to run this notebook you'll need to install pymc3 + link to install guide.

For this notebook you'll also need to install bambi+link to bambi.

This could also go in a separate hidden cell (so it's not visible in website but it's visible in notebook)
:::

:::{dropdown} Suggested reading
:icon: book 

This notebook assumes you are familiar with x, y and z. If you are not, consider reading notebook on x, notebook on y... before reading this notebook.
```

OriolAbril:
We could also use substitutions to keep the extra content in the notebooks to a minimum: https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#substitutions-with-jinja2

OriolAbril:
idea (needs to be tested): provide some templates for installation instructions, could have a paragraph of intro, be inside a dropdown, have tabs within the dropdown for pip/conda installs... but leave the packages to be installed as jinja substitutions. We can then add notebook metadata to define the substitutions, see https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#substitutions-with-jinja2.

main expected problem: install commands should probably be formatted as code but I don't know how to get jinja to work inside code.

drbenvincent:
Is this issue now defunct/solved given we have the [Style Guide](https://docs.pymc.io/en/latest/contributing/jupyter_style.html)?

PR #239, opened 2021-10-14
improve appearance

### LINKED ISSUE #235: Show tags and categories within notebook posts
Currently tags and categories are seen when searching for a notebook (i.e. from [tag pages](https://pymc-examples.readthedocs.io/en/latest/blog/tag/pymc3model.html)), but not once the notebook is [open](https://pymc-examples.readthedocs.io/en/latest/time_series/MvGaussianRandomWalk_demo.html). Might be natively supported by [ablog](https://ablog.readthedocs.io/en/latest/)

OriolAbril:
Work on the overall appearance of the website.

* Add "view on github", "launch on binder" and "open with collab" badges to all **posts**. Only
  notebooks with the post directive will have this. Partial resolution of https://github.com/pymc-devs/pymc-examples/issues/200, and it is automated.
    * binder env still needs to be configured though.
* Add postcard information to the left sidebar. This will include tags and categories (thus this fixes #235) but it could
  be interesting to include authors too as discussed in #198.
    * The current proposal is a bit clunky but might work well enough. It does have a bit of repeated info though which is not ideal. I have basically hijacked the ablog template to include the authors with a link to `#authors` which should then be set and contain the authorship info, links to the profiles of authors...
* ~Make the sidebar scrollable, see https://github.com/pydata/pydata-sphinx-theme/issues/500~ This will soon be fixed directly by the theme in a much more pretty way than what I first did here.
* Fix logos and PyMC3 instances in sphinx config; the notebooks and tags still need updating.


OriolAbril:
Here is how the spline example looks like for example: https://pymc-examples--239.org.readthedocs.build/en/239/splines/spline.html which is where I have added authors to the post directive to show how it looks like cc @tjburch

martinacantaro:
This is not a hard preference, but I would prefer authorship to be at the end of the notebook, and in line with [this proposal](https://github.com/pymc-devs/pymc-examples/issues/198#issuecomment-916486451). I like it better because it has more nuance than displaying the names of everybody who made a contribution, which would put someone who fixed a typo at the same level of someone who created the entire thing. 

I think this is useful for readers in cases where they want to know who did the main work for a notebook (for example, if they need to quote the notebook, or if they want to get in touch with the author). 

Also, I think for the purposes of documentation it's not as important to display authorship prominently, so I would leave the sidebar for categories and tags.

OriolAbril:
todos: 

* [ ] update the requirements to codeautolink released version, not github
* [ ] add doi link and code to html_config once we have them. There is an issue with the webhook right now and they have not been generated even after my fake releases.
* [ ] (for someone else who knows pre-commit to do) Add a pre-commit check to ensure all notebooks end with the include page_footer directive and if possible add it automatically when missing.

OriolAbril:
Now realized a bit of info on codeautolink is probably in order. It is a sphinx extension that automatically converts the python objects from the libraries configured on intersphinx-mappings to links to their relevant api pages. It is very new and still needs some tweaks, but it is active and mostly works.

Here are some examples form their docs: https://sphinx-codeautolink.readthedocs.io/en/latest/examples.html

OriolAbril:
Ready to merge and I'll merge this weekend even without any other approving review.

martinacantaro (approved):

examples/splines/spline.ipynb
```diff
@@ -4,16 +4,12 @@
    "cell_type": "markdown",
```
michaelosthege, line 4: <pre class="ql-syntax" spellcheck="false">that are tied togehter at their boundaries
                  ☝
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/239/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

michaelosthege (approved):
Found a typo that was there before.

Also the spline notebook still uses "PyMC3", not sure if you want to change that already.

OriolAbril:
> Also the spline notebook still uses "PyMC3", not sure if you want to change that already.

We'll need to rerun them all with v4, so we can take care of that then. So far I have not updated that anywhere in notebooks, only in readme and templates.

Issue #47, opened 2021-03-24
Multilevel modeling


OriolAbril:
file: https://github.com/pymc-devs/pymc-examples/blob/main/examples/case_studies/multilevel_modeling.ipynb

This notebook endured a significant rewrite not too long ago, and is one of the go to examples on ArviZ+xarray usage. This is both its main pro and its own downfall. There are some nits to keep it up to its current golden standard.

## Required changes
The notebook should be reexecuted with latest pymc3 and have the warning filter removed (there should be no warnings, if there are they should be fixed). Moreover, rerunning with latest ArviZ will generate much more aesthetically pleasing forestplots. `legend=True` should be used for forestplots with multiple models.

## Changes for discussion
The plot forest on cell 42 should use the approach described in https://github.com/arviz-devs/arviz/issues/1627 to avoid showing the 1 dots that correspond to the elements in the diagonal.

:warning: These changes need to wait for the next ArviZ release, they require ArviZ>0.11.2, 0.11.2 is not enough.

farhanreynaldo:
Hi @OriolAbril, does this issue also have a time constraint? If not, I would like to work on this issue.

OriolAbril:
This one also has the high impact label, but needs very little changes. Only re executing it with latest pymc and arviz and some cosmetic changes to follow the style guide: https://github.com/pymc-devs/pymc/wiki/PyMC-Jupyter-Notebook-Style-Guide

farhanreynaldo:
Ah sure, may I assign myself on this issue? I will try my best to make the PR as soon as possible.

PR #246, opened 2021-11-09
update multilevel modeling notebook

farhanreynaldo:
Updated the multilevel modeling notebook with the following improvements:

* Abide by code preamble.
* Try-except to load data.
* Run on the latest PyMC and Arviz.
Addresses issue https://github.com/pymc-devs/pymc-examples/issues/47.

Need to figure why the layout constrained warning is popped out and `covariation_intercept_slope` yielded Overflow Error on the latest library.


farhanreynaldo:
<p>I've tried both approaches, and it showed a practically similar figure. I think we could remove the <code>tight_layout</code>.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/246/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

farhanreynaldo:
<blockquote>The kde can' be generated for constants and it has never been, but somehow we have changed from silently failing and plotting nothing without warning to erroring out.</blockquote><p> </p><p>That's make sense. I will ignore the diagonal of chol_corr in this case.</p><p> </p><p>And also when I opened this issue in Arviz, are you suggesting proposing warning and no plot when the constant cases happened? </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/246/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
<p>sounds great</p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/246/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
<p>a bit like it's done with tags. The spline notebook does it: https://github.com/pymc-devs/pymc-examples/blob/main/examples/splines/spline.ipynb. The metadata we add in this post directive is hidden in the rendered version, but we can then use it for example to generate a recommended citation as we are setting up in https://github.com/pymc-devs/pymc-examples/pull/239</p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/246/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

farhanreynaldo:
<p> </p><p>Sure, I've updated it on the latest commit.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/246/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

farhanreynaldo:
<p>Understood, I have updated it in the latest commit. Let me know if I missed anything.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/246/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
Thanks for the great work so far, left one last comment before merging

farhanreynaldo:
<blockquote>This is also skipping one of the elements of chol_stds. I think we need to do two plots, one for all vars like right now but adding <code>chol_corr</code> to the variables to skip, then in the cell below plotting chol_corr only.</blockquote><p> </p><p>Don't you think it would be confusing for the readers if we separate the plot between the <code>chol_corr</code>  and the rest of the vars? We do this due to the limitation of current implementation, right?</p><p> </p><blockquote>Here it is an overkill, but I think it's good for illustrative purposes, as already in 3d we'd need to use that.</blockquote><p> </p><p>I am relatively new and unfamiliar with the xarray, but from my current understanding the array for<code>covariation_intercept_slope_idata.posterior['chol_corr']</code>  is 2-dimensional, as shown below:</p><pre class="ql-syntax" spellcheck="false">array([[[[ 1.        , -0.28202288],
         [-0.28202288,  1.        ]],

        [[ 1.        , -0.73111316],
         [-0.73111316,  1.        ]],
        ...
</pre><p>Did I understand it incorrectly? </p><p> </p><p>And also if we are going to proceed to separate the plots, does the following code is the correct implementation?</p><p> </p><pre class="ql-syntax" spellcheck="false"># without chol_corr
az.plot_trace(
    covariation_intercept_slope_idata,
    var_names=["~z", "~chol", "~chol_corr"],
    compact=True,
    chain_prop={"ls": "-"},
);
</pre><p>and the next cell is </p><pre class="ql-syntax" spellcheck="false">az.plot_trace(
    covariation_intercept_slope_idata,
    var_names=["chol_corr"],
    lines=[("chol_corr", {}, 0.0)],
    compact=True,
    chain_prop={"ls": "-"},
    coords={"param": ["a"], "param_bis": ["b"]},
);
</pre>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/246/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

farhanreynaldo:
<p>Or do you mean adding <code>coords={"param": ["a"], "param_bis": ["b"]}</code>  would also affected the <code>county</code>  and <code>chol_stds</code>  variables?</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/246/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
<p>Yes, it is two dimensional which is why it works in this case, but it only works in 2d which is not always the case, and I think we should use the example from the issue _even if it's not necessary here_ so that the notebook is useful also for people who might have &gt;=3d corr matrices.</p><p> </p><blockquote>Or do you mean adding <code>coords={"param": ["a"], "param_bis": ["b"]}</code> would also affected the <code>county</code> and <code>chol_stds</code> variables?</blockquote><p> </p><p>Yes, it has modified the output of the chol_stds plot. coords selection is applied to the whole dataset and therefore it affects all variables that have those dimensions (so county is not really affected because it's multidim but with other dimensions).</p><p> </p><p>The code example you shared with the two cells is the idea I was proposing. The coords need to be updated to use this pointwise selection (again, not necessary here because we are only in 2d but needed for 3 or more dims). You can also use <code>var_names="chol_corr"</code>  the list is no necessary.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/246/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
changes look great, but the notebook doesn't look executed now, can you make sure it is executed with the outputs stored in the ipynb file? The preview looks like this: https://pymc-examples--246.org.readthedocs.build/en/246/case_studies/multilevel_modeling.html

farhanreynaldo:
> changes look great, but the notebook doesn't look executed now, can you make sure it is executed with the outputs stored in the ipynb file? The preview looks like this: https://pymc-examples--246.org.readthedocs.build/en/246/case_studies/multilevel_modeling.html

My bad, I forgot to rerun the notebook, I am on it.

farhanreynaldo:
Should I also add the epilogue section in this notebook?

OriolAbril:
> Should I also add the epilogue section in this notebook?

no, we might move this one to pymc again in order to maintain a handful of base notebooks always synced with the latest releases and versioned, so those might look a bit different than the rest of examples.

OriolAbril (approved):

farhanreynaldo:
Thank you so much for your guidance for this PR @OriolAbril!

PR #257, opened 2021-12-17
Bayesian copula estimation example notebook

drbenvincent:
This pull request adds a new example notebook for Bayesian copula estimation. Work done by myself and @ericmjl .

Currently works under v3. Not working under v4, possible because `LKJCholeskyCov` is not refactored for v4 yet (according to @ricardoV94)


OriolAbril (commented):
I don't know why the readthedocs build is failing, will try to look into it at some point.

drbenvincent:
<p>Thanks. Have done these things now.</p><p> </p><p>I also simplified the logos. Rather than 2 separate files in a markdown table, I just have a single image with both logos now and no markdown table. Should avoid any strange table rendering issues.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>This should hopefully be fixed now.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Thanks, I didn't know SciPy distributions took <code>rng</code> , but now I do :) Fixed</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>I'm not sure I follow this question. Is this about lines 14, 15, 16? That computation could be done before, outside of the model as <code>a_mu</code>,  <code>a_sigma</code> , <code>b_sigma</code>  are basically observed variables. But I'm not sure that's what you are getting at.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Might need input from Eric here to double check</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>As far as I can tell, we could use <code>copula_idata.posterior.cov</code>  rather than calculate ppc samples. </p><p> </p><p>Doing this results in <code>copula_idata.posterior.cov.stack(sample=("chain", "draw")).data.shape</code>    being 2x2x4000. So we then need to iterate over this last <code>sample</code>  dimension rather than doing <code>dists = [multivariate_normal([0, 0], cov) for cov in ppc["cov"]]</code>  </p><p> </p><p>But it's not immediately obvious to me how to iterate over the <code>sample</code>  dimension of an xarray object. Any suggestions?</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
Thanks @OriolAbril. I've dealt with all of those points except for the last one which I have a question about.

FYI. Also need some review (when @ericmjl is able) before we merge this. Could you tag his as a reviewer? I don't seem to be able to request reviewers.

drbenvincent:
I don't understand why the graphviz images aren't displaying.

ericmjl:
<p>Some notes from talking with Oriol:</p><p> </p><ol><li>The entire block of deterministics can be moved out of the model context.</li><li>We'll probably want to explain why we're using point estimates.</li></ol><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ericmjl:
<p>Possibly a transpose. (Just leaving a note from calling Oriol.)</p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
<p>The <code>:class: myclass</code>  is an example to show extra attributes can be used, we don't need to include it here</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
<p>The ref won't work like this. reviewnb messed up the formatting, but you need 3 things, the reference type (ref), the text to be shown (here) and the target where the link should point to (awkward_binning). See https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#the-ref-role for more details</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
> I don't understand why the graphviz images aren't displaying.

Don't trust reviewnb preview, always check the readthedocs preview to see if things look good or not. reviewnb messes up quite often, and even in the case it did work, what readers will actually see is readthedocs, so this is what we need to double check.

ericmjl:
<p>@drbenvincent I think a transpose is what we can do. For an array that is of shape (2, 2, 4000), I think a transpose using <code>np.transpose(cov, axes=(2, 0, 1))</code>  should do the trick. I think that syntax will take the position 2 dimension and move it to position 0.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
<p>I did some xarray+arviz black magic and updated all the calculations to use the full posterior, which does seem to give the same result. Feel free to use all or parts of that. If doing the transposition here however, do it with xarray and take the values after that. You can use <code>copula_idata.posterior.cov.stack(sample=("chain", "draw")).transpose("sample", ...)</code>  (the ... being an actual ellipsis) to ensure the sample dimension is the first and the others keep the same order</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
Thanks @OriolAbril. That last commit cleans everything up, and uses your recommended way of getting the posterior data of `cov` out in a convenient format for plotting and comparison to the data. I've also added you to the acknowledgements 🙂

ericmjl:
<p>I think the spelling of transform is incorrect. Should be <code>transform_data</code>.</p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Fixed</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Typo fixed.</p><p> </p><p>I will add some text (shortly) to mention why we are using point estimates.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Extra comma now removed.</p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>In fact, we've already got a note about point estimates under the heading "PyMC models for copula and marginal estimation"</p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
I _think_ that should've addressed remaining concerns. But happy to do more minor things if there's something I've missed.

OriolAbril:
> In fact, we've already got a note about point estimates under the heading "PyMC models for copula and marginal estimation"

The note is about simultaneous estimation or fitting marginal and covariance models independently, but doesn't address marginalizing before or after the transformation between observed and multivariate spaces. The marginalization after transforming is one if the things I tried and did give the same result, but the two approaches giving the same result isn't straightforward (to me at least) as the operations don't commute with the expectation.

I would also change this comment

> data munging to extract covariance estimates from copula_idata in ~useful shape~ different format

Now it's a bit easier to follow because we only draw one sample per posterior draw from the multivariate normal, but I needed to run the original code that drew 100 pp samples per posterior draw to understand what was happening. The xarray heavy code is longer and has some extra lines that should not be necessary but allows to read input shapes and how they are modified.

We should probably abstain from judgment calls but especially about xarray being mostly annoying, which I think many people believe already. It is perfectly understandable that this is the case but given how it is becoming the default output we should try to not give that impression ourselves (while we note and work to make xarray easier to work with for stats or linalg postprocessing which are currently quite annoying having to call apply_ufunc manually)

ericmjl:
<p>Copying Oriol's point from the discussion to keep the discussion a bit easier to follow:</p><p> </p><blockquote>The note is about simultaneous estimation or fitting marginal and covariance models independently, but doesn't address marginalizing before or after the transformation between observed and multivariate spaces. The marginalization after transforming is one if the things I tried and did give the same result, but the two approaches giving the same result isn't straightforward (to me at least) as the operations don't commute with the expectation.</blockquote><p> </p><p>I think I see where Oriol's point - if the expectation of a distribution is E(dist), then exp(E(dist)) is not necessarily equal to E(exp(dist)). </p><p> </p><p>However, we're doing an exp(logcdf) here, not exp(E(dist)). Oriol, I'll readily admit to being ignorant here, but does that change anything?</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
The thing is that even if the evaluation of the logcdf are the observations and thus not a random variable, it's parameters are, so I think it's something like E(probit(exp(logcdf(x|dist)))) or probit(exp(logcdf(x|E(dist)))) which doesn't seem like they must be the same

ericmjl:
<p>Copying Oriol's point from the GitHub PR tracker to keep the discussion easier to follow:</p><p> </p><blockquote>The thing is that even if the evaluation of the logcdf are the observations and thus not a random variable, it's parameters are, so I think it's something like E(probit(exp(logcdf(x|dist)))) or probit(exp(logcdf(x|E(dist)))) which doesn't seem like they must be the same</blockquote><p> </p><p>Upon thinking about your point, Oriol, I think it is easiest for now to leave a not on why we're not using the whole posterior. I think I get your point, and it makes sense, though I'm also aware that we should close things out before they get dragged on too long, and it's easiest to add a comment than to try to change the logic of the code.</p><p> </p><p>Ben, please be on the lookout for a few more textual changes to push!</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/257/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
For the record... this PR is not dead. It is on my list to deal with, but has been low priority.

drbenvincent:
Housekeeping done. Ready to try to get this over the finish line...

drbenvincent:
Took me a while to fix some issues - waiting for docs to build etc. But notebook looks good now :)

twiecki (approved):

examples/case_studies/binning.myst.md
```diff
@@ -10,7 +10,7 @@ kernelspec:
   name: conda-env-pymc_env-py
 ---
 
-(awkward_binning)=
+(binning)=
```
OriolAbril, line 13: we should keep both anchors, I don't know if the original one is already being used somewhere. We did something similar in https://github.com/pymc-devs/pymc-examples/blob/main/examples/generalized_linear_models/multilevel_modeling.ipynb when we combined two notebooks in one.

OriolAbril (changes requested):
requesting one change to make sure we don't break any cross-references, after that ready to merge

examples/case_studies/binning.myst.md
drbenvincent: I did it anyway 😂

PR #259, opened 2021-12-21
Add BART introductory example

aloctavodia:
This notebook includes two examples showing how to use BART in PyMC and also some intuition about how BART works.  I tried to keep it light on the technical details related to BART and focus more on its practical features. But I am opne to add  further details explaining BART itself or maybe regression trees or whatever you think is necessary to improve the notebook.

Some references are still missing (I will add them soon as part of this PR), at least the seminal bart paper should be linked and also our upcoming CRC press has a chapter about BART (although with the old implementation). 

Also I am working with others on an ArxiV preprint that will work as "extended documentation" of BART with more technical details of our particular implementation of bart and it sampler and also more examples to show strength and limitations of BART and the PyMC implementation. Also in future PRs I will like to add one or two more BART notebooks  showing other uses and examples.


examples/BART/BART_introduction.ipynb
```diff
@@ -0,0 +1,557 @@
+{
```
OriolAbril, line 1: <p>You need to add a target at the top of the notebook too so it can be referenced from other notebooks or the versioned docs: https://docs.pymc.io/en/latest/contributing/jupyter_style.html#first-cell</p><p> </p><p>I think the notebook is clear and beginner friendly, and would change the category to beginner. It feels a bit strange to have introduction in the title but be intermediate. I ended up commenting only on formatting and one code issue.</p><p> </p><p>This is still in early discussion but we were considering with Martina to use 2 categories for notebooks, level and type one. The type one would be one of the 4 types defined in diataxis. This notebook imo fits quite well the explanation type: https://diataxis.fr/explanation/#the-language-of-explanation. Given that you are planning more notebooks on the topic, would that division and structure help? Do you think it would be useful for users?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/259/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

OriolAbril, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.    rng = np.random.RandomState(RANDOM_SEED)
</pre><p><code>np.random.default_rng</code> the RandomState is the legacy generator. According to numpy docs:</p><p> </p><blockquote>This class should only be used if it is essential to have randoms that are identical to what would have been produced by previous versions of NumPy.</blockquote><p> </p><p>reference: https://numpy.org/doc/stable/reference/random/legacy.html#legacy</p><p> </p><p>note: if we don't generate samples with numpy we can have only a random seed passed to the pymc.Model instances</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/259/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

OriolAbril, line 1: <p>you are probably planning that already, but the <code>1</code>  and <code>2</code>  should be references: https://docs.pymc.io/en/latest/contributing/jupyter_style.html#references</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/259/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

OriolAbril, line 1: <p>coal mining link is broken</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/259/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

OriolAbril, line 1: <p>Is this in the released beta version or the development pymc one? Those warning need to be fixed because otherwise the code will break in 1-2 ArviZ releases.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/259/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

OriolAbril, line 1: <p>Add yourself as author above: https://docs.pymc.io/en/latest/contributing/jupyter_style.html#authorship-and-attribution</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/259/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

OriolAbril, line 1: <p>add the epiloge markdown cell: https://docs.pymc.io/en/latest/contributing/jupyter_style.html#epilogue</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/259/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

aloctavodia: Thanks for the feedback.

I am not sure about the criteria used for beginner, intermediate or advance categories. For example this notebook is and introduction to BART but assumed reader already knows Bayesian statistics and PyMC. So is beginners with respect to BART but non beginners with respect to Bayesian modeling.

Having more structure will give us a little bit more of work. Buy I think that will be beneficial for users, and even for us, that way it will be easier to spot missing documentations or areas than need more attention to get a balanced documentation. Also structure is very helpful for the autodidactic, that not only needs to read some specific documentation to solve a particular problem but want to use the documentation as part of a large learning plan.

aloctavodia: using `np.random.default_rng` does not work with `pm.Model()` I can use an integer for that as a workaround as you suggested, but still  it will not work with `pm.bart.predict`. So I guess this is ok for know until we update pymc to work with the new generator?

aloctavodia: Now I am confused, is ok to use markdown links for example to point to Wikipedia or other webpages? Or we should always use latex/bibtex style? For example here `1` and `2` are links to web pages not scientific papers. `Poisson processes` and `Cox processes` are links to Wikipedia articles

aloctavodia: is this the correct way? [coal mining](../getting_started.ipynb)

aloctavodia: <p>I run the notebook from main. There are a couple of tweaks in <code>m.bart.plot_dependence</code>  not available in the released beta version.</p>

OriolAbril: Not so sure either myself about those levels either, we should probably define them somewhere (cc @martinacantaro) and not worry about those too much until then.

If you find the diataxis structure useful, add `explanation` as second category (and then choose one category and label the other posts in the same manner). Several parts of the versioned docs are heavily inspired by diataxis so I think it will be good to try to follow it everywhere (to some extent)

OriolAbril: Yeah, didn't know the update on pymc side was still missing. Is there an issue I cpuld subscribe to to avoid asking people to change this when it's not yet possible?

OriolAbril: Wikipedia links and to other doc pages are fine, yes. I commented because I opened one and looked at it for a whole 2 seconds and got the impression it was a vignette that would then have a citation with the package. No need to do it if it's not the case

OriolAbril: Nope, use `` {ref}`pymc:pymc_overview` ``. It might not work yet but will once the repevant PR in pymc updating the notebook is merged. As this is rebuilt with every commit it should not take too long for everything to work without further action on our side

OriolAbril: I will then add a label to wait for beta2 to rerun it then merge. Notebooks should run on the env shown in the watermark which would not be the case

aloctavodia: Not sure, @ricardoV94 probably knows.

examples/BART/BART_introduction.ipynb
```diff
@@ -0,0 +1,576 @@
...
```
OriolAbril, line 1: <p>the <code>(notebook_id)=</code>  is the identifier we'll then use from other notebooks to link here. It should be something specific to this notebook. I'd use the filename.</p><p> </p><p>The 2nd category should be like with tags, comma separated instead of a second category metadata key: <code>:category: intermediate, explanation</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/259/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

aloctavodia:
Given that we have a new release this should be ready to merge.

canyon289 (approved):

OriolAbril:
It needs to be executed with the new release though. Otherwise the watermark will say beta1 and this notebook won't run on beta1, only >=beta2

OriolAbril (approved):

PR #262, opened 2022-01-07
Search-and-replace a bunch of strings to port to v4.

twiecki:
Here is the script I ran:
```bash
#!/bin/bash

find ./ -name "*.ipynb" -type f -exec sed -i -e 's/az.from_pymc3/pm.to_inference_data/g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/arviz.from_pymc3/pm.to_inference_data/g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/pymc3/pymc/g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/PyMC3/PyMC/g' {} \;
find ./ -name "*.py" -type f -exec sed -i -e 's/pymc3/pymc/g' {} \;
find ./ -name "*.py" -type f -exec sed -i -e 's/PyMC3/PyMC/g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/from theano import tensor as tt/import aesara.tensor as at/g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/import theano\.tensor as tt/import aesara.tensor as at/g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/tt\./at./g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/aet/at/g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/studenat/studentt/g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/theano/aesara/g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/Theano/Aesara/g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/pm\.sample()/pm.sample(return_inferencedata=False)/g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/, return_inferencedata\=True//g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/return_inferencedata\=True, //g' {} \;
find ./ -name "*.ipynb" -type f -exec sed -i -e 's/return_inferencedata\=True//g' {} \;
pre-commit run --all-files
```

We probably need a new branch for v4 here.

Issue #52, opened 2021-03-29
BEST


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/case_studies/BEST.ipynb
Reviewers: @twiecki @fonnesbeck

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
* Use numpy [Generator](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator) instead of using the global state random state. Extra reference: https://numpy.org/doc/stable/reference/random/index.html?highlight=random%20sampling%20numpy%20random#quick-start

### Exotic dependencies
None

### Computing requirements
Models sample in less than a minute

larryshamalama:
Submitted a PR [here](https://github.com/pymc-devs/pymc-examples/pull/92)!

asuagar:
Hello @OriolAbril! I would like to try to update this notebook to PyMC 4. I have done BEST before using NumPyro. I have a dumb question about the graphical representation. Can it be used seaborn or other one?

OriolAbril:
Assigned it to you @asuagar, thanks for helping out!

> I have a dumb question about the graphical representation where ArviZ is not used. Can it be used seaborn or other graphical library?

Other libraries can be used for plotting, in many cases ArviZ is not the right tool for the job. i.e. ArviZ is designed to explore bayesian model results, so preliminary exploratory analysis should generally use something else, then use ArviZ afterwards to analyze the posterior. Was that the question?

asuagar:
Yes, it was. It is now clear to me. After reading several notebooks, I thought that there was some kind of limitation to Matplotlib for exploratory data analysis. Thank you by your fast reply.

PR #264, opened 2022-01-08
Add rerun script

michaelosthege:
This PR adds `scripts/rerun.py` and applies it to the BEST notebook.
It can be used from the command line or from another Python script to apply API migrations, re-run notebooks and optionally commit to a specific branch and push it to a remote.

The replacements are the ones @twiecki wrote for #262.

For example, this is how we can apply it for the example scripts:
```
import rerun

for fp in ["example1.py", "example2.py"]:
    assert rerun.apply_replacements("some_example.py")
    assert rerun.commit(fp, "update-some-example")
assert rerun.push("update-some-example", "myremotename")
```

@twiecki @OriolAbril I think we should chunk the work and, for example, have one PR that updates only the example scripts like shown above.
If needed those PRs can then adapt the `REPLACEMENTS`.


scripts/rerun.py
```diff
@@ -0,0 +1,156 @@
...
+import logging
+import pathlib
+import subprocess
+
+
+logging.basicConfig(level=logging.INFO)
+_log = logging.getLogger(__file__)
+DP_REPO = pathlib.Path(__file__).absolute().parent.parent
+
+REPLACEMENTS = {
```
twiecki, line 27: The order matters here.

michaelosthege: Ah, so that's why there's the `stundenat`→`studentt`!
I though you were correcting typos there..

OriolAbril:
This is related to #52, so we keep track of everything. I will also create a `v4 (auto)` column in the notebook tracker to take into account that they have been updated but might still need code updated to show v4 best practices, loops of scalar distributions converted to correct use of the size argument, removing now unnnecessary model factories things like that. 

The PR LGTM, and merging this and aiming for multiple PRs also allows anyone to help.

OriolAbril (approved):

OriolAbril:
#265 updates the BEST example on both code and style, can we autoupdate another notebook here to avoid merge issues?

twiecki:
What happened to the AB testing NB?

michaelosthege:
> What happened to the AB testing NB?

I undid that commit to avoid conflicts, as requested by
https://github.com/pymc-devs/pymc-examples/pull/264#issuecomment-1008267679

The notebook I re-ran now needed a manual intervention, but I got annoyed by automated re-runs not working, so I just fixed it.
Most breaking changes that need manual intervention happen around working with `InferenceData` instead of `MultiTrace`.

OriolAbril:
related to #110

scripts/rerun.py
twiecki: Should we not make this an OrderedDict or can we rely on `dict` being ordered now?

scripts/rerun.py
```diff
@@ -0,0 +1,171 @@
...
+    "az.from_pymc3": "pm.to_inference_data",
+    "arviz.from_pymc3": "pm.to_inference_data",
+    "pymc3": "pymc",
+    "PyMC3": "PyMC",
+    "pymc3": "pymc",
```
twiecki, line 32: these occur twice.

PR #274, opened 2022-01-30
Updated splines notebook to v4

fonnesbeck:
Modest updates to convert to v4.
- replaced `PyMC3` with `PyMC` througout
- now uses RNG seeding via `Model` instantiation
- few corrections and updates to text


michaelosthege:
<p>Yes, the <code>Model(coords=...)</code>  are only <code>np.arange</code>  and those would be auto-generated.</p><p>You can write this:</p><pre class="ql-syntax" spellcheck="false">w&nbsp;= pm.Normal("w", mu=0, sd=3, size=B.shape[1], dims="splines")
</pre><p>to automatically create a lenght-3 dimension tied to that <code>(B.shape[1],)</code>-sized RV.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/274/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
<p>it does and it's standard code formatting. The strange view is all because of reviewnb. The actual change is swapping apostrophes for backticks. Never trust reviewnb preview. What matters is what is shown in the website: https://pymc--274.org.readthedocs.build/projects/examples/en/274/splines/spline.html?highlight=splines#prepare-the-spline</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/274/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

fonnesbeck:
Should be good to go now.

michaelosthege (approved):

myst_nbs/splines/spline.myst.md
```diff
@@ -6,17 +6,17 @@ jupytext:
...
   language: python
   name: python3
 ---
 
 # Splines in PyMC3
 
-:::{post} Oct 8, 2021 
-:tags: patsy, pymc3.Deterministic, pymc3.Exponential, pymc3.Model, pymc3.Normal, regression, spline
+:::{post} May 6, 2022 
+:tags: patsy, pymc.Deterministic, pymc.Exponential, pymc.Model, pymc.Normal, regression, spline
```
OriolAbril, line 17: There should not be `pymc.<object>` tags anymore, that generates way too many tags that are not that useful so we decided to use https://docs.pymc.io/projects/examples/en/latest/object_index/index.html instead for this. The issue to track progress on this is https://github.com/pymc-devs/pymc-examples/issues/289

myst_nbs/splines/spline.myst.md
```diff
@@ -6,17 +6,17 @@ jupytext:
...
 :category: beginner
-:author: Joshua Cook, Tyler James Burch
+:author: Joshua Cook updated by Tyler James Burch, Chris Fonnesbeck
```
OriolAbril, line 19: This is metadata used to generate the citation advise at the bottom. It should only have names and exclude those who have only re-executed the notebooks without making changes. The Authors section should always be present and list everyone and the changes they did.

We discussed that both in an issue, in the docs channel in slack and in a couple doc meetings. Like I often say, I do not have strong feelings about what was chosen and would have been perfectly happy (or even more happy) with another choice or with changing this now. What I do have strong feelings about is that all notebooks do the same because otherwise some people don't dare add their name at all whereas others do which only increases already present inequalities. Therefore changing that would mean changing the style guide and updating all the notebooks that are already in the [Done](https://github.com/pymc-devs/pymc-examples/projects/1) column.

PR #278, opened 2022-02-04
try setting up pre-commit hook for bibtex

OriolAbril:
@MarcoGorelli is there a way to require pre-commit to use a minimum version of node?

This library needs v12 or newer, and while I have been able to update node locally and run the library from the command line, we should not ask contributors to install/update node only to run pre-commit.


MarcoGorelli:
[language version](https://pre-commit.com/#overriding-language-version) might work for this?

OriolAbril:
Everything is working now, thanks! :heart:

PR #279, opened 2022-02-05
Update Jupyter style in lasso notebook

ltoniazzi:
# Addressing [Issue #110](https://github.com/pymc-devs/pymc-examples/issues/110)
Upgrading with respect to the style guides and v4 practices to reach Best Practices. Main changes:
- Changed some variables names to follow notation y = X * beta
- Add myst references of some pymc classes (still need to fix them as they are not rendering correctly)
- header, footer, authors...
- removed `start` in `pm.sample`.

I am not sure if changes should be done with respect to v4. Can someone comment on that? @fonnesbeck @CloudChaoszero  @michaelosthege 

### Helpful links


examples/pymc3_howto/lasso_block_update.ipynb
```diff
@@ -4,40 +4,75 @@
    "cell_type": "markdown",
```
OriolAbril, line 4: <p>The <code>notebook_id</code>  is the key we'll use from other notebooks (or versioned docs) to link here, it should be somehting notebook specific like the filename. I will update the advise to say it must be the filename, you are at least the 4th person confused by that.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/279/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

OriolAbril, line 4: <p>the formatting doesn't look right. It should be <code>{class}'pymc.sample'</code>  but with backticks instead of apostrophes. <strong>However</strong>, important note, it won't render right! This is an issue with pymc versioned documentation that has been fixed for v4 but the notebooks try to generate links to stable (and this is fixed in latest only)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/279/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

OriolAbril, line 4: <pre class="ql-syntax" spellcheck="false">Line #2.        initvals = pm.find_MAP()
</pre><p>it might be best to get rid of this completely instead of updating to use initvals. Is there any reason to use the map as initial point?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/279/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

OriolAbril, line 4: <p>pylab should not be used anymore as a general rule, pyplot can be used in its place. However, in this particular case, I think it's better to use <code>plot_pair</code> , see example in https://arviz-devs.github.io/arviz/examples/plot_joint.html (it uses plot_pair, not plot_joint). Feel free to set marginals to True or to False here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/279/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

ltoniazzi: I did, however as I side note, once I remove `initvals`, sometimes it runs correctly but also many times the sampling hangs and I get some warnings. Specifically, running:
```
with model:
    step1 = pm.Metropolis([beta1, beta2])
    step2 = pm.Slice([sigma, tau])
    idata = pm.sample(10000, step=[step1, step2])
```
I sometimes get three of these RuntimeWarnings
```
/home/user/miniconda3/envs/pymc-test-py39/lib/python3.9/site-packages/pymc/step_methods/metropolis.py:250: RuntimeWarning: overflow encountered in exp
  "accept": np.exp(accept),
```
and the sampling hangs.

OriolAbril: Does this happen during tuning or during sampling?

michaelosthege: <pre class="ql-syntax" spellcheck="false">`"accept": np.exp(accept)
</pre><p>`that's probably an extreme value in the acceptance probability during MCMC. Since you're seeing this go away with MAP initivals, my best guess would be that the prior is far away from the typical set.</p><p>Looking at the traceplot below, that's most probably somethin with the "tau" parameter. The default initval for this U(0,1) variable will be 0.5, but the posterior is <em>very</em> close to 0.</p>

ltoniazzi: Thanks for the hints. The issue happens during sampling. Also, the lam scaling is equivalent to the original notebook, and I tried changing it to lam=1, leave only tau=HalfNormal(0,1) and some other combinations, but I still get this issue on most runs without initvals. You think I should put initvals back? Also, if I run pm.sample(10000) (without steps and initvals) everything works fine.

OriolAbril (approved):

Issue #216, opened 2021-08-30
Hierarchical NB has missing images


twiecki:
https://docs.pymc.io/pymc-examples/examples/generalized_linear_models/GLM-hierarchical.html

Missing are the images of the hierarchical model that @fonnesbeck created. Not sure if he has an updated link.

saiwing-yeung:
Same for https://github.com/pymc-devs/pymc-examples/blob/main/examples/case_studies/multilevel_modeling.ipynb
in the `## Multilevel and hierarchical models` cell.

drbenvincent:
This is not fixed yet.

New notebook location with new website: https://github.com/pymc-devs/pymc-examples/blob/main/examples/generalized_linear_models/GLM-hierarchical.ipynb

Still needs to be updates to best practices.

PR #282, opened 2022-02-13
Have pooling model images be local as they keep disappearing.

twiecki:
Fixes https://github.com/pymc-devs/pymc-examples/issues/216.


OriolAbril (changes requested):
Super in favour of adding the images locally, but they should still use the markdown image syntax (which works for both remote and local images) as the html version is parsed as raw content by sphinx, so the images are not copied correctly when generating the html pages, you can see that the preview still has broken images: https://pymc--282.org.readthedocs.build/projects/examples/en/282/case_studies/multilevel_modeling.html#multilevel-and-hierarchical-models

canyon289:
@twiecki FYI theres some bug on github viewer for NB images. @AlexAndorra was running into the same thing

cfonnesbeck:
Working on this now. Will push a fix today.

twiecki:
@fonnesbeck is this superseded by #285?

PR #283, opened 2022-02-16
add myst notebook representation of examples

OriolAbril:
As we start taking advantage of more and more myst features, the reviewnb representation of
the notebooks becomes more and more limited. I think we should keep it because it is useful
to see a diff with outputs, but the content diffs are often wrong (i.e. if adding backticks
for code formatting), tags and metadata is not shown anywhere (and we use that for myst
substitutions, hiding cells, rendering code generated images as figures...).

Having a myst representation of the notebooks solves this problem. I am also adding
jupytext as a pre-commit hook to make sure ipynb and .myst.md files always stay synced.


OriolAbril:
Discussed that with @martinacantaro and @drbenvincent in the docs meeting and decided to go ahead with this.

PR #287, opened 2022-03-08
Update LKJ notebook to v4

fonnesbeck:
Cleanup and update to v4 and Aesara


cfonnesbeck:
Not sure what the jupytext failure entails. There is just a lot of embedded image garbage, which makes it hard to diagnose.

AlexAndorra (approved):
LGTM, thanks @fonnesbeck

OriolAbril:
Sent a PR to fix CI.

> Not sure what the jupytext failure entails.

Notebooks have now each 2 files, a ipynb one and a myst one. The jupytext pre-commit keeps them synced, you only need to care and work on the ipynb one when writing/executing notebooks. When reviewing however, the ReviewNB has too many blind spots, and the myst view is very useful; it shows the text changes much more clearly. Reviewnb messes up big time when the text is not changed but it's formatting is and starts showing html tags as if they were part of the diff, it ignores the breaklines (as it should when rendering markdown, only while lines are relevant, but with myst when using directives breaklines are important) and also ignores completely all the tags and metadata of the notebooks which can be used to hide code cells or outputs that are interesting but not key to the notebook, to add captions and/or alt text to code generated plots... I shared the link to #283 which added this syncing and we also discussed that in a documentation meeting where we decided the extra pre-commit was worth it to make sure we can review easily even when using all these features I mentioned.

PR #298, opened 2022-03-20
Simpsons paradox update

drbenvincent:
(see #213)

Am attempting to update this notebook to best practice. Should be ok, but right now I'm having some shape issues when using `pm.set_data` for posterior predictive inference.

Any pointers here appreciated... I don't have a good mental model of what `dim`'s should be provided (or not). Is there a definitive reference of `coords` and what one needs to consider when using `pm.set_data`?


examples/generalized_linear_models/GLM-truncated-censored-regression.ipynb
```diff
@@ -955,7 +955,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.7"
+   "version": "3.9.9"
```
OriolAbril, line 958: this file should be excluded from the PR

myst_nbs/generalized_linear_models/GLM-simpsons-paradox.myst.md
```diff
@@ -6,13 +6,21 @@ jupytext:
...
-  name: python3
+  name: pymc-dev-py39
 ---
 
+(GLM-simpsons-paradox)=
 # Simpson's paradox and mixed models
-Author: [Benjamin T. Vincent](https://github.com/drbenvincent)
+
+:::{post} March, 2022
+:tags: regression, hierarchical model, linear model, multi level model, posterior predictive
```
OriolAbril, line 18: what are the differences between hierarchical and multi model?

myst_nbs/generalized_linear_models/GLM-simpsons-paradox.myst.md
```diff
@@ -524,7 +535,20 @@ The third and final model added a layer to the hierarchy, which captures our kno
 
 If you are interested in learning more, there are a number of other [PyMC examples](http://docs.pymc.io/nb_examples/index.html) covering hierarchical modelling and regression topics.
```
OriolAbril, line 536: this link needs to be updated. I guess the idea is to point to the homepage? if so you can use `` {doc}`index` ``

OriolAbril (commented):
to avoid sync issues it might be best to address the comments below in the notebook directly

drbenvincent:
<p>I think there is just the one core problem that I'm having here, and after figuring that out everything else will be easy. I tried your suggestion, but this does not update <code>idata</code>  with a predictions block - so it seems to not update in place.</p><p> </p><p>Using <code>ppc = pm.sample_posterior_predictive(idata, predictions=True)</code> works (kind of). </p><ul><li><code>ppc.predictions_constant_data.x</code>  will be of the right size</li><li><code>ppc.predictions.y</code>  is not the right size, (chain: 4, draw: 1000, y_dim_0: 100) </li></ul><p> </p><p><code>linear_regression.x.eval().shape</code> shows that <code>pm.set_data</code> <em>is</em> working, but I don't understand why posterior predictive samples of x has 100 observations (the original data size) and not the updated size of 20.</p><p> </p><p>I think we are seeing an example of your concern here... the code is moving faster than the docs... I'm not aware of any documentation that really goes over current best practice for coords and changing data with set_data. Although your various comments here are very useful.   </p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/298/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
Aside from pre-commit failing (I assume because these were updated recently), I think this is ready. But let me know if you spot anything.

OriolAbril (approved):
Approving, but given that CI here is short, it would be best to rebase after merging https://github.com/pymc-devs/pymc-examples/pull/304. Feel free to merge the PR yourself once CI is green

Note that after merging you should close the tracker issue (which will automatically move the card to the "Done" column in the [tracker project](https://github.com/pymc-devs/pymc-examples/projects/1))

Issue #275, opened 2022-01-30
Merge Cox survival example with Bayesian survival analysis example


fonnesbeck:
While the Cox survival notebook is a nice example of a ported BUGS model, there is no supporting documentation in the notebook. Moreover, there is already a Cox example in Austin's comprehensive Bayesian survival analysis notebook. I propose 3 alternatives:

1. Integrate the Cox example into the Bayesian survival analysis notebook
2. Turn the Cox example into a `.py` script
3. Flesh out the Cox notebook with supporting narrative

I'm not personally in favor of the third option, as it is a lot of work for what is mostly a redundant example, but I want to get feedback before choosing one and updating.

OriolAbril:
I would also argue against option 2, python files are too hard to find and I think we should probably see which ones to move into notebooks and get rid of the others. 

Given that the survival analysis notebook is already quite long, I'd like to propose a different option, splitting the notebook (in two or more) with a notebook being an [explanation](https://diataxis.fr/explanation/) about survival analysis and different approaches and the others [how to guides](https://diataxis.fr/how-to-guides/) on writing a specific type of survival model with pymc, one of them being the cox one? Note that I haven't gone through either notebook carefully and I'm not sure how feasible this proposal is.

cfonnesbeck:
We have several non-notebook examples already. One (not me) could argue that not everyone likes examples in notebook format. If we don't want additional python scripts, then I agree we should be converting/removing the existing ones.

There's nothing inherently wrong with a long notebook, particularly for a complicated topic like survival modeling. This is analogous to vignettes in the R world.  I don't think integrating the Cox notebook would make it much longer, as there is already a Cox model in there; it would mostly be cherry-picking the good parts to modify the existing one.

OriolAbril:
> We have several non-notebook examples already. One (not me) could argue that not everyone likes examples in notebook format. If we don't want additional python scripts, then I agree we should be converting/removing the existing ones.

in my opinion, any python/jupyter differences are completely eclipsed by the fact that python scripts are not on the website. GitHub shows traffic for top pages for the last 14 days. The last page to make it today is the BART PR, with a total of 11 views (from 3 unique visitors only). No pyhton script makes the list. Given how we have things set up now, python script examples are as good as gone.

> There's nothing inherently wrong with a long notebook

[...]

PR #306, opened 2022-04-09
Remove cox example

drbenvincent:
After discussions (#275) we are removing the Cox proportional hazard example (see also #127).


OriolAbril:
The myst one needs to be removed too, not sure why the pre-commit check is not working on deletion :/

OriolAbril (approved):
can you open an issue so we don't forget there is something wrong with the pre-commit check?

drbenvincent:
done, see #307.

Ok for me to merge?

Issue #292, opened 2022-03-09
Add a pre-commit check to prevent urls pointing to pymc docs


OriolAbril:
It should be the same as https://github.com/pymc-devs/pymc/pull/5486 but here in pymc-examples.

PR #311, opened 2022-04-14
[ENH] Add a pre-commit check to prevent urls pointing to pymc docs

juanitorduz:
Fixes https://github.com/pymc-devs/pymc-examples/issues/292

I would support by solving this issue. However from the description I am not 100% the cope of which subfolders I need to add. I am taking the implementation from https://github.com/pymc-devs/pymc/pull/5486 as suggested in the issue description. 

Once I get some initial feedback I will keep working on this one :)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,66 +1,66 @@
 repos:
-  - repo: https://github.com/mwouts/jupytext
-    rev: v1.13.7
-    hooks:
-      - id: jupytext
-        files: ^examples/.+\.ipynb$
-        args: ["--sync"]
-  - repo: https://github.com/psf/black
-    rev: 22.3.0
-    hooks:
-      - id: black-jupyter
-  - repo: https://github.com/nbQA-dev/nbQA
-    rev: 1.1.0
-    hooks:
-      - id: nbqa-isort
-        additional_dependencies: [isort==5.6.4]
-      - id: nbqa-pyupgrade
-        additional_dependencies: [pyupgrade==2.7.4]
-        args: [--py37-plus]
-  - repo: https://github.com/MarcoGorelli/madforhooks
-    rev: 0.3.0
-    hooks:
-      - id: check-execution-order
-        args: [--strict]
-        exclude: |
-          (?x)^
-          ^examples/ode_models/ODE_with_manual_gradients\.ipynb$
-          |examples/samplers/DEMetropolisZ_EfficiencyComparison\.ipynb$
-          |examples/gaussian_processes/GP-Latent\.ipynb$
-          |examples/gaussian_processes/GP-MaunaLoa2\.ipynb$
-          |examples/samplers/MLDA_gravity_surveying\.ipynb$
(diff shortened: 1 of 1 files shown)


OriolAbril:
Thanks! It should check all the files inside the `examples/` folder except the `conf.py` file.

juanitorduz:
@OriolAbril is https://github.com/pymc-devs/pymc-examples/pull/311/commits/dd3c1380b071bafca56e8c22e3d891b59cbf850f enough or should add / remove websites from the `entry: >` specification?

.pre-commit-config.yaml
```diff
@@ -55,6 +55,27 @@ repos:
       minimum_pre_commit_version: 2.8.0
       name: Check notebooks have watermark (see Jupyter style guide from PyMC docs)
       types: [jupyter]
+    - id: no-internal-links
+      name: Check no internal links are in the docs
+      description: >-
+        'A quick check to prevent urls pointing to pymc docs'
+      files: ^examples/
+      exclude: >
```
OriolAbril, line 63: The 404.md file should also be excluded. And for now I would exclude all the notebooks that fail the check too so we can merge the PR and then work on fixing the issues

.pre-commit-config.yaml
```diff
@@ -55,6 +55,27 @@ repos:
...
+          (?x)(conf.py)
+      entry: >
+          (?x)(arviz-devs.github.io|
+               aesara.readthedocs.io|
+               aeppl.readthedocs.io|
+               pymc-experimental.readthedocs.io|
+               docs.pymc.io|
+               numpy.org/doc|
+               pymc-examples.readthedocs.io|
+               myst-parser.readthedocs.io|
```
OriolAbril, line 73: The two myst ones can be removed, but `python.arviz.org` and `docs.xarray.dev` should be added

juanitorduz: Added in https://github.com/pymc-devs/pymc-examples/pull/311/commits/22e83a6e6c6f1b3f7e8ac70c767422268420b6af

.pre-commit-config.yaml
juanitorduz: `404.md` added in https://github.com/pymc-devs/pymc-examples/pull/311/commits/22e83a6e6c6f1b3f7e8ac70c767422268420b6af

juanitorduz: I added all the notebooks which were failing and I also had to add `index.md` to that the tests pass.

OriolAbril:
Thanks! Can you check if any of the now temporarily excluded notebooks is also in the "Done" column of the tracker project? https://github.com/pymc-devs/pymc-examples/projects/1

juanitorduz:
Out of the 13 tickets in the "Done" column I do not see any which is also in this exclusion list.

OriolAbril (approved):
> Out of the 13 tickets in the "Done" column I do not see any which is also in this exclusion list.

Great, thanks for the PR, I think it is ready to merge

PR #323, opened 2022-05-20
update BART example

aloctavodia:
Updates the notebook to reflect the last changes in pymc-experimental, including the new plot for variable importance/selection.


examples/BART/BART_introduction.ipynb
```diff
@@ -24,7 +24,7 @@
      "name": "stdout",
```
OriolAbril, line 24: <p>you should remove all the <code>pymc.</code> tags. We will use a different functionality to find notebooks that use pymc functions and classes instead of tags, and leave the tags for "conceptual" topics only. Also, when it comes to the tags, I would choose either the acronym BART or the full name and use only one of them. For barts using the full name might be better</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/323/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

OriolAbril, line 24: <p>Bit of a side comment, also related to the discussion on discourse. The function signature has <code>X=None</code>,making it an optional parameter, but it doesn't look like it's optional  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/323/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

aloctavodia: Right, I will fix this.

aloctavodia: <p>Just to be clear. I have to remove the tags "pymc3.*" but keep others like "BART"?</p>

OriolAbril (approved):

Issue #160, opened 2021-05-20
Examples gallery thumbnails are sometimes poorly representative


JohnGoertz:
It seems that the [examples gallery](https://docs.pymc.io/nb_examples/index.html) auto-generates thumbnails by grabbing the last figure in a notebook. However, this has often resulted in thumbnails that are either un-informative or even mis-informative. For example, the thumbnail for GP-circular is actually a contrasting example the author provided of how *not* to make a circular GP. In a similar vein, the "Mauna Loa continued" thumbnail is not particularly informative.

Perhaps the notebooks could be tweaked to simply put a more useful figure last, or thumbnails could be specified via a cell metadata tag in the notebooks, or a thumbnail could be uploaded separately from the notebook?

OriolAbril:
having no idea how thumbnail generation works, the metadata option sounds good. I think it was @ColCarroll who wrote the sphinx extension to add example to the gallery so he may have better insight

PR #328, opened 2022-05-26
Gallery

### LINKED ISSUE #327: Implement nbsphinx gallery
We would like to bring the NB gallery from v3 back

The old code was customly written, which makes maintenance harder.

`nbsphinx` allows to include NBs in sphinx and has a gallery feature that is described here: https://nbsphinx.readthedocs.io/en/0.8.8/subdir/gallery.html

twiecki:
Using a hack to only use the parts of `nbsphinx` that are required for the gallery but does not clash with `myst_nb`.

Closes #327.

Related: https://github.com/executablebooks/MyST-NB/issues/396
Hack from: https://github.com/PlasmaPy/PlasmaPy/pull/1474#issuecomment-1062476158

I couldn't yet figure out how to get it to work with `md` files rather than `rst`. I also couldn't figure out how to make it look only for `*-ipynb` files, it only worked with `*`.

@OriolAbril do you have an idea?

image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,7 +45,6 @@ repos:
     - id: bibtex-tidy
       files: examples/references.bib
       args: ["--sort=key", --quiet]
-      language_version: 15.14.0
 - repo: local
   hooks:
     - id: watermark
--- a/examples/conf.py
+++ b/examples/conf.py
@@ -1,4 +1,4 @@
-import os, sys
+import os
 from sphinx.application import Sphinx
 
 # -- Project information -----------------------------------------------------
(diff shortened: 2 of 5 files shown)


OriolAbril:
This looks promising, thanks!

> I couldn't yet figure out how to get it to work with md files rather than rst.

Using rst for this page is not an issue, the notebooks continue to be jupyter notebooks which is the important part.

> I also couldn't figure out how to make it look only for *-ipynb files, it only worked with *.

We can look into that, but it might be easier to just get rid of the python scripts mixed in there. Python scripts don't appear on the website nor make the list in the github traffic section which means that at most they get 5 visits a month. Otherwise they would appear in this traffic page. If some script is very important is should be made into a notebook, otherwise it is as if it didn't exist and will end up being removed at some point.

I'll try to take a better look next week and see if we can get the images to work and maybe also 4 columns in the thumbnails

twiecki:
Also allows to address https://github.com/pymc-devs/pymc-examples/issues/160.

PR #340, opened 2022-05-31
Get thumbnails working for gallery

twiecki:
Hackish, but it works.

Ideally we could use `nbsphinx` functionality to extract the thumbnails via the cell metadata but I couldn't get it to work (ref https://github.com/spatialaudio/nbsphinx/issues/648).

This instead does things manually: it uses the old code @ColCarroll wrote to extract images from NBs, saves them to a dir, and then tells `nbsphinx` to reference against them.

image

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,5 +7,3 @@ python:
    version: "3.8"
    install:
    - requirements: requirements-docs.txt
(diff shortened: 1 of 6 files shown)


twiecki:
Alright, rebased!

Issue #346, opened 2022-06-01
Thumbnails in gallery not working


twiecki:
https://github.com/pymc-devs/pymc-examples/pull/340 should have fixed it, but on https://www.pymc.io/projects/examples/en/latest/gallery.html thumbnails are not showing up. Although it is working locally.

twiecki:
It seems like the thumbnail generator script isn't called.

PR #347, opened 2022-06-01
Refactor fix thumbnails.

twiecki:
Rename gallery_generator to thumbnail_extractor. Add debug output. Write thumbnails to separate directory.

Fixes https://github.com/pymc-devs/pymc-examples/issues/346.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 
 _build
 jupyter_execute
+_thumbnails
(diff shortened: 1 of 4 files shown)


twiecki:
OK, thumbnails are showing up now on the build docs.

OriolAbril (approved):
skimmed the changes and they look good, the only comment is I think we should add the thumbnail directory to the gitignore.

I won't be able to join today, and the navbar PR will need to be rebased after this I think. It would be great if someone can take that over and merge it so the website already looks the same for all sites.

PR #352, opened 2022-06-01
update rolling regression notebook to v4

drbenvincent:
see #89

This should be pretty much best practice now.

Only meaningful change I made was with one plot because `pm.plot_posterior_predictive_glm` no longer exists


drbenvincent:
Can't make much sense of the failing test. Something to do with `az.plot_hdi`?

fonnesbeck (approved):

PR #354, opened 2022-06-01
Update LGCP notebook to v4

fonnesbeck:
In addition to updating to v4, I removed the second model from the notebook. The first model already takes upward of an hour to run on a reasonably fast PC; the second model takes almost a day.

The notebook has also been moved from the case studies folder to the gaussian processes folder.


twiecki:
NB looks good, but pre-commit. Probably need to update myst file.

OriolAbril (approved):

Issue #353, opened 2022-06-01
Gallery page is missing howtos, tutorials, and deep-dives


drbenvincent:
As far as I can tell, none of the notebooks in [`examples/pymc3_howto`](https://github.com/pymc-devs/pymc-examples/tree/main/examples/pymc3_howto) folder are currently included/rendered in the examples page https://www.pymc.io/projects/examples/en/latest/gallery.html. They _are_ there when looking for them via the tags, but they are not rendered on the gallery page.

twiecki:
Yeah, just need to add them, it's quite simple.

PR #359, opened 2022-06-02
Fix gallery

drbenvincent:
Addresses #353 

Changes:
- renamed folder `pymc3_howto` to `pymc_howto`
- added folders to `gallery.rst`

Feedback wanted on:
- order of items
- are there any folders we _don't_ want to show for some reason that I'm not aware of?


OriolAbril (commented):
looks good, you should also delete the pymc3_howto folder inside myst_nbs folders

drbenvincent:
As far as I can tell the failures might be coming from:
- cells not in order in `sampling_callback.ipynb`
- `Check no internal links are in the doc` fails, but unclear what notebook! Maybe `examples/pymc_howto/api_quickstart.ipynb`

Can we still consider merging and deal with those issues in separate pull requests?

drbenvincent:
In there built docs, what is going on with the BART section?!

twiecki (dismissed):

twiecki:
Can you rename `pymc_howto` just to `howto`?

twiecki:
We shouldn't break CI.

drbenvincent:
> Yeah I'd just rerun the callback_nb.

Well that is still pymc3. Could either upgrade or run in an environment with v3

drbenvincent:
> That happens when there's an error, like the same name appearing twice.

That is in fact the problem. Fix incoming...

drbenvincent:
do not merge yet

drbenvincent:
Ok, so checks pass.

TOC view in gallery is fixed for some, but still getting it for BART and splines. No idea why the problem remains for this.

examples/gallery.rst
```diff
@@ -48,6 +48,16 @@ Gaussian Processes
...
 
+BART
+----
+
+.. nbgallery::
+    :name: BART
+    :glob:
+    :reversed:
+
+    BART/*
```
twiecki, line 59: Isn't BART moved to experimental anyway?

OriolAbril: yes, but examples with experimental should also be here in principle

drbenvincent:
Done. Can I get an approval @twiecki or @OriolAbril, then this is ready to merge 🎉

PR #361, opened 2022-06-02
Fix gallery

drbenvincent:
Further to #353, this PR removes left over headings in the gallery.

The single BART and Spline examples were put into Case Studies to avoid a gallery rendering error, but I forgot to remove the sections from the `gallery.rst`.

So this PR fixes that and should enable #353 to be closed.


drbenvincent:
Working correctly EXCEPT FOR no gallery thumbnails under How To right at the bottom

examples/gallery.rst
```diff
@@ -124,16 +114,4 @@ How to
 .. nbgallery::
     :name: howto
     :glob:
-    :reversed:
-
-    howto/*
```
OriolAbril, line 129: This doesn't look right and the how tos are gone in the preview: https://pymcio--361.org.readthedocs.build/projects/examples/en/361/gallery.html#how-to

drbenvincent: Just noticed lack of thumbnails. What is wrong with it?

OriolAbril: I think you deleted the `howto/*` with the latest merge from main somehow and so nothing is listed inside that "nbgallery"

drbenvincent: Ok. Will come at this again tomorrow with a fresh head.

OriolAbril (approved):

Issue #58, opened 2021-03-29
probabilistic matrix factorization


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/case_studies/probabilistic_matrix_factorization.ipynb
Reviewers: @colcarroll

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
* Use numpy [Generator](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator). See also https://numpy.org/doc/stable/reference/random/index.html?highlight=random%20sampling%20numpy%20random#quick-start

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### ArviZ related
* Use ArviZ and xarray for postprocessing. This will probably be challenging. I'd recommend familiarizing with xarray before working on that. Some ideas:
   * `_norms` in code cell 23 looks like it could be replaced by `xr.apply_ufunc` (using `input_core_dims`)
### Exotic dependencies
None

### Computing requirements
Model samples in roughly 1 hour

OriolAbril:
Note: the "good first issue" label only applies to updating the notebook from `To Do` to `General Updates`.

chiral-carbon:
Hi @OriolAbril , can I take this issue up? I am going through the probabilistic matrix factorization notebook at the moment and thought I could attempt to make the recommended changes.

OriolAbril:
Definitely,  I'll assign it to you. Let me know if you have any doubts, especially on xarray which can be challenging even for seasoned pymc3 users

chiral-carbon:
sure, thanks! I'll get back to you on this.

PR #365, opened 2022-06-03
Probabilistic Matrix Factorization for Making Personalized Recommendations to v4

percevalve:
# Moving to v4 of Probabilistic Matrix Factorization for Making Personalized Recommendations

Related Issue: #58 
Also remove the default to one chain in `draw_samples` (`kwargs.setdefault("chains", 1)`)

+ [X] PR description contains a link to the relevant issue: a tracker one for existing notebooks or a proposal one for new notebooks
+ [X ] Check the notebook is not excluded from any pre-commit check: https://github.com/pymc-devs/pymc-examples/blob/main/.pre-commit-config.yaml

### Helpful links


drbenvincent:
If the conflicts can be resolved, and the suggested tags added, then this is probably done?

percevalve:
Indeed, that should be done.
The only difference I noted was a different result with find_map, but it does not change the meaning and the interpretation of the results.

drbenvincent:
> Indeed, that should be done.

So it's saying that there are conflicts which need to be resolve before this can be approved and merged. But it's a bit strange because the conflicting files are the ones being updated. Any insights @OriolAbril?

drbenvincent (approved):

percevalve:
Let me just re-run it to finish the NB and add the tag.

drbenvincent:
I think this is ready. But it won't let me merge until the conflicts are resolved. I'm a bit confused as it is the particular files that are being changed which are conflicting, and I thought this was the point. Any insights @percevalve or @OriolAbril ?

percevalve:
Looks like I solve the conflict, but the version is not the latest run... running it now.

Issue #128, opened 2021-04-05
survival analysis


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/survival_analysis/survival_analysis.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
* fix typos
* fix math renderend incorrectly
* update notation to use equal tailed interval (or ETI) instead of HPD (or HDI), both are valid credible intervals, but they are different, so we should not write HPD if using ETI

### ArviZ related
* Use arviz (or only import? I don't really understand how that notebook runs)
* Use InferenceData and xarray for exploratory analysis of the results

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

olgadk7:
In addition, should I update the notebook as mentioned in #4027?

fonnesbeck:
@olgadk7 That would be terrific, thanks!

OriolAbril:
That would be amazing indeed, thanks! Just to clarify, we don't expect you to fix all the bullet points above, choose which ones you want to address and let us know so we can review based on that and force extra work on you after reviewing.

olgadk7:
Regarding ArviZ, in the survival_analysis notebook it doesn’t do anything that pymc3 couldn’t, does it make sense then to only use pymc3 perhaps? Unless I’m missing some rules that govern how to chose between the two (didn’t find this info after a quick search, just that you seemingly always need to set the ArviZ style, as per the Notebook Style Guide).

Related to that, I’m confused how and why the notebook’s source code https://github.com/pymc-devs/pymc-examples/blob/main/examples/survival_analysis/survival_analysis.ipynb calls to az. (without even importing it), but the published notebook https://docs.pymc.io/notebooks/survival_analysis.html uses pymc3 only. 

On another note, Can I add the “with model” context argument to get rid of the corresponding FutureWarning of deprecation?

[...]

PR #372, opened 2022-06-05
Adapted code for survival_analysis.ipynb to work with v4

cuchoi:
Updates survival analysis notebook to v4. Issue #128.

+ [x] PR description contains a link to the relevant issue: a tracker one for existing notebooks or a proposal one for new notebooks

### Helpful links


cuchoi:
The code (new and old) has a very high number of divergences (>100). Would it be worth it to try to reduce that? Although the r_hat values look good, it seems that we shouldn't be encouraging analysis with this level of divergences?

canyon289:
If you can do so but otherwise add a note with a disclaimer that this
amount of divergences would warrant further investigation. it's a good
teaching opportunity!

cuchoi:
Makes sense! I think I will add a disclaimer since I don't know enough about these type of models to get rid of the divergences

canyon289:
Sounds good, and if youd like open up an issue ticket as well so someone else can come by later and try their hand at reducing the divergences

twiecki:
Thanks @cuchoi!

Issue #385, opened 2022-06-30
Pre-commit getting in the way


drbenvincent:
I’m trying to do an initial commit on my fork of `pymc-examples` so I can do a PR of a new notebook. 

I run the pre-commit checks with `pre-commit run --all`  and it all passes. But then when I try to commit I get errors complaining that the ipynb and myst files are inconsistent. So I delete the myst file and regenerate it by running `pre-commit run --all` again, and I get the same errors. So I’m unable to commit.

Would be great to find a solution to this particular issues, which is just one of a number of specific problems in contributing to the repo.

drbenvincent:
One of the problems I am facing is that the local pre-commit checks are changing the spelling in a URL (which is a not sensible thing to be trying to do). Depending on what way the wind is blowing it seems to only do this on the myst files which then creates a conflict between the ipynb and myst files.

Long story short, the use of `codespell` in the precommit code is causing real issues.

This seems to be introduced by #321 by @hectormz.

Is there a way to get `codespell` to NOT try to fix spelling in URLs? This seems like a very odd thing to be doing in the first place.

If we can't do that, can we just get rid of `codespell` and get people to do this manually? It's making things extremely difficult and is sapping all the fun out of contributing to this repo

hectormz:
@drbenvincent it looks like you have a solution in #388 that might do the trick. I agree, if `codespell` is frustrating or getting in the way, something should be changed.

PR #387, opened 2022-07-01
NEW NOTEBOOK - counterfactual inference via calculating excess deaths during Covid-19 era

drbenvincent:
Hi. This PR contributes a new notebook. The primary topic is calculating excess deaths in England and Wales during the Covid-19 era. The main things of interest are the concepts covered:
- causal inference
- linear regression for time-series data
- Bayesian workflow

Links to Google's Causal Uplift approach are drawn.

Very happy to receive any feedback 🙂

NOTE: it is likely that the checks will fail. I have been having a _very hard time_ with the pre-commit checks. As far as I can work out, there is some auto spelling modification which is applied only to the myst version of the notebook which causes a conflict between the ipynb and myst files. I do not know how to get around this as I have very little experience with commit hooks. I was only able to even commit by bypassing the pre-commit checks with `git commit --no-verify`.


cluhmann:
It looks like black is reformatting a couple of lines in the notebook at which point the notebook is no long consistent with the jupytext/myst file.  If you can either run black before committing or manually format the notebook so that black doesn't want to make any changes (and then save so that the myst gets updated), does that help?  You can see what lines diverge in the failed pre-commit.

cluhmann:
I would also suggest separating out the binning notebook into a separate PR.

drbenvincent:
> I would also suggest separating out the binning notebook into a separate PR.

I'll look into this, I didn't realise I committed any changes to this. It shouldn't be there

drbenvincent:
> I would also suggest separating out the binning notebook into a separate PR.

Done. No idea why changes to the binning files got wrapped up in this

drbenvincent:
Tests pass! It's a miracle!

Although it has autocorrected a URL from `www.ons.gov.uk/` to `www.owns.gov.uk` Why are we getting autocorrection on URLs? That seems like madness.

drbenvincent:
Right. This is not going to pass unless I purposefully include the wrong URL, which is not sensible. We need to do something about `codespell` in the precommit hooks, see #385.

cluhmann:
Can you try using the full URL?  That is, "http://" whatever?  [This](https://github.com/codespell-project/codespell/pull/1656/files) seems so suggest that URLs should be excluded from spell-checking.

drbenvincent:
> Can you try using the full URL? That is, "http://" whatever? [This](https://github.com/codespell-project/codespell/pull/1656/files) seems so suggest that URLs should be excluded from spell-checking.

Odd, I'm already doing that:
Screenshot 2022-07-02 at 19 32 09

cluhmann:
Or perhaps that's an option that can be enabled on codespell?  I didn't look closely.

drbenvincent:
Looks like there are some args:

https://github.com/pymc-devs/pymc-examples/blob/9de026c86192f312b919958c79e44b90328c8a63/.pre-commit-config.yaml#L106-L111

cluhmann:
So maybe the solution to the codespell part of #385 is to modify that pre-commit.

drbenvincent:
Yep. Trying to figure out what the correct argument is

drbenvincent:
Think I've got it. I'll create a separate PR to fix the codespell URI stuff

drbenvincent:
Ok. I've filed a separate PR here https://github.com/pymc-devs/pymc-examples/pull/388 to fix this. 

I initially tried to just edit the precommit file locally. This then does make the local pre-commit checks pass BUT it was not letting me push the changes while the `.pre-commit-config.yaml` file is unstaged.

So in order to make progress here, we first have to close #388. Then I can rebase or whatever and make progress with this notebook.

cluhmann:
Yeah, if the only thing holding up this PR is some pre-commit stuff then I think that's probably fine.  If we get this PR approved, it will still be blocked until the pre-commit gets fixed.  But then this will be automatically unblocked once that happens.  So now I guess I can take a look at the actual notebook!

drbenvincent:
> Yeah, if the only thing holding up this PR is some pre-commit stuff then I think that's probably fine. If we get this PR approved, it will still be blocked until the pre-commit gets fixed. But then this will be automatically unblocked once that happens. So now I guess I can take a look at the actual notebook!

Happy to get comments on the content. I believe @juanitorduz will have time to provide additional review next week - we're back to needing 2 approving reviewers.

drbenvincent:
Thanks for everyone's input on this so far. Had some issues with my main branch so life will be simpler switching over to this new PR, #391 
I'll close this PR momentarily and do the changes in the new PR

drbenvincent:
> It would be great to have a follow up notebook where we allow more complex models like a local linear trend and maybe auto-regressive terms? :)

Yes, I think they could be idea for the future. Perhaps for a more time-series focussed notebook.

PR #388, opened 2022-07-02
Update `.pre-commit-config.yaml` to tell `codespell` to ignore spelling in URI's

drbenvincent:
Fixes a frustration (see #385) where the pre-commit checks are editing the spelling within a URL to something which is wrong. It also has a complex set of effects on the ipynb vs myst files which can create differences and making it impossible to commit anything.

This PR updates `.pre-commit-config.yaml` to tell `codespell` to ignore spelling in URI's simply by appending `"--uri-ignore-words-list", "*"` to the arguments for `codespell``

PR #390, opened 2022-07-04
Fix codespell - third time lucky

drbenvincent:
3rd attempt to get this right from #388 

> Fixes a frustration (see #385) where the pre-commit checks are editing the spelling within a URL to something which is wrong. It also has a complex set of effects on the ipynb vs myst files which can create differences and making it impossible to commit anything.
> 
> This PR updates `.pre-commit-config.yaml` to tell `codespell` to ignore spelling in URI's simply by appending `"--uri-ignore-words-list", "*"` to the arguments for `codespell``

Hopefully I've correctly rebased my main. This PR _should_ only have one change to `.pre-commit-config.yml`


drbenvincent:
What the hell is going on?

Issue #272, opened 2022-01-14
Reinforcement Learning Model for behavioral data


ricardoV94:
# Notebook proposal
**Title**: Reinforcement Learning Model for behavioral data

## Why should this notebook be added to pymc-examples?

I wrote this notebook some time ago: https://github.com/ricardoV94/stats/blob/master/modelling/RL_PyMC.ipynb and shared it on the Discourse, where it seems to have been quite popular: https://discourse.pymc.io/t/fitting-a-simple-reinforcement-learning-model-to-behavioral-data-with-pymc3-jupyter-nb/4651

## Related notebooks
I don't know any related notebooks...

# References
Hopefully the base notebook is good enough. I can provide more references if necessary

***

## Notes
I have been meaning to adapt this into an official pymc-examples notebook but I simply can't find the time. I am happy to assist anyone who wants to work on this, and I have no issues if the final version modifies/extends/trims it or if it doesn't resemble the original at all!

The original code was written in Theano, but I think everything will work fine with the equivalent Aesara imports.

juanitorduz:
@ricardoV94 I need to learn about this topic and your notebook is a hidden gem ;) I have seen your last update (commit) was a month ago. Will you push this notebook to the finish line or do you need some support? I'll be happy to help. Great work (as always)!

ricardoV94:
@juanitorduz If you mean adding it to pymc-examples, it's up for grabs. I don't plan to work on it in the foreseeable future.

If you want to extend it, many people have asked how to include multiple participants/experiments and not just a single one as in the example.

juanitorduz:
Yes, put it into the docs! I suggest (as I am not an expert in the subject) format the notebook in order to publish it in the docs , get feedback and iterate.

ricardoV94:
> Yes, put it into the docs! I suggest (as I am not an expert in the subject) format the notebook in order to publish it in the docs , get feedback and iterate. 

That would be great!

PR #410, opened 2022-08-04
Reinforcement Learning Notebook

juanitorduz:
Closes https://github.com/pymc-devs/pymc-examples/issues/272

+ [x] PR description contains a link to the relevant issue: a tracker one for existing notebooks or a proposal one for new notebooks

### Helpful links


myst_nbs/case_studies/reinforcement_learning.myst.md
```diff
@@ -0,0 +1,550 @@
...
+
+$$ P(a = \text{right}) = \frac{\exp(\beta Q_{\text{right}})}{\exp(\beta Q_{\text{right}}) + \exp(\beta Q_{\text{left}})}$$
+
+where the $\beta \in (0, +\infty)$ parameter determines the level of noise in the agent choices. Larger values will be associated with more deterministic choices and smaller values with increasingly random choices.
+
+***
+
+### Credits
+* The PyMC code was adapted from that of Maria Eckstein ([GitHub](https://github.com/MariaEckstein/SLCN), [PyMC Discourse](https://discourse.pymc.io/t/modeling-reinforcement-learning-of-human-participant-using-pymc3/1735))
+* The MLE code was adapted from that of Robert Wilson and Anne Collins {cite:p}`collinswilson2019` ([GitHub](https://github.com/AnneCollins/TenSimpleRulesModeling))
```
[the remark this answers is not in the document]
juanitorduz: ok! This is absolutely 100% Ricardo's work. I am just fixing the code style and pushing it to make it more visible via the pymc docs.

myst_nbs/case_studies/reinforcement_learning.myst.md
```diff
@@ -0,0 +1,550 @@
...
+
+    plt.ylabel("action")
+    plt.xlabel("trial")
+
+    handles, labels = plt.gca().get_legend_handles_labels()
+    order = (1, 2, 0)
+    handles = [handles[idx] for idx in order]
+    labels = [labels[idx] for idx in order]
+
+    plt.legend(handles, labels, fontsize=12, loc=(1.01, 0.27))
```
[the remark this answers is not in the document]
juanitorduz: Thanks for these comments! I think is a great addition to the notebook. I will add them.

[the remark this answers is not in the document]
juanitorduz: So...I added the plot suggestion in https://github.com/pymc-devs/pymc-examples/pull/410/commits/69acb1f64770f06b1098e9b62114f1780f817b8d but I was not able to use `Line2D` to create the legend as you suggested so for now this is something where I would appreciate some tips.

myst_nbs/case_studies/reinforcement_learning.myst.md
```diff
@@ -0,0 +1,550 @@
...
+kernelspec:
+  display_name: Python 3.9.13 ('website_projects-1IZj_WTw')
+  language: python
+  name: python3
+---
+
++++ {"id": "Pq7u0kdRwDje"}
+
+(reinforcement_learning)=
+# Fitting a simple Reinforcement Learning model to behavioral data with PyMC
```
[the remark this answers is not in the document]
juanitorduz: @ricardoV94 any suggestion on the title? (for now I removed the "simple" in https://github.com/pymc-devs/pymc-examples/pull/410/commits/69acb1f64770f06b1098e9b62114f1780f817b8d)

myst_nbs/case_studies/reinforcement_learning.myst.md
```diff
@@ -0,0 +1,550 @@
...
+
+```{code-cell} ipython3
+:id: SDJN2w117eox
+
+az.plot_posterior(data=tr_alt, ref_val=[true_alpha, true_beta]);
+```
+
+```{code-cell} ipython3
+%load_ext watermark
+%watermark -n -u -v -iv -w -p aesara,pymc
```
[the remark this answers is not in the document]
juanitorduz: Done in https://github.com/pymc-devs/pymc-examples/pull/410/commits/69acb1f64770f06b1098e9b62114f1780f817b8d

myst_nbs/case_studies/reinforcement_learning.myst.md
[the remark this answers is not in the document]
juanitorduz: Also, I was not able to hide the plot code cell. AFAIK I need to use `:tags: [hide-cell]` in the myst.md file? but then the pre-commit complains because the notebook and the `myst.md` are not consistent ... can you please give me a tip here :) ?

myst_nbs/case_studies/reinforcement_learning.myst.md
```diff
@@ -0,0 +1,541 @@
...
+ax.legend(handles, labels, fontsize=12, loc=(1.01, 0.27));
+```
+
++++ {"id": "6RNLAtqDXgG_"}
+
+The plot above shows a simulated run of 150 trials, with parameters $\alpha = .5$ and $\beta = 5$, and constant reward probabilities of $.4$ and $.6$ for the left (blue) and right (orange) actions, respectively. 
+
+Solid and empty dots indicate actions followed by rewards and no-rewards, respectively. The solid line shows the estimated $Q$ value for each action centered around the respective colored dots (the line is above its dots when the respective $Q$ value is above $.5$, and below otherwise). It can be seen that this value increases with rewards (solid dots) and decreases with non-rewards (empty dots). 
+
+The change in line height following each outcome is directly related to the $\alpha$ pamater. The influence of the $\beta$ parameter is more difficult to grasp, but one way to think about it is that the higher its value, the higher the probability of flipping between actions regardless of their estimated values.
```
ricardoV94, line 170: There's a mistake in this paragraph. The higher the `beta` the more deterministic the responses are / or the more sensitive they are to the observed rewards. I think flipping the adjective does not quite convey this information...

```suggestion
The change in line height following each outcome is directly related to the $\alpha$ parameter. The influence of the $\beta$ parameter is more difficult to grasp, but one way to think about it is that the higher its value, the more an agent will stick to the action that has the highest estimated value, even if the difference between the two is quite small. Conversely, as this value approaches zero, the agent will start picking randomly between the two actions, regardless of their estimated values.
```

myst_nbs/case_studies/reinforcement_learning.myst.md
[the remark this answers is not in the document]
OriolAbril: > So...I added the plot suggestion in https://github.com/pymc-devs/pymc-examples/commit/69acb1f64770f06b1098e9b62114f1780f817b8d but I was not able to use Line2D to create the legend as you suggested so for now this is something where I would appreciate some tips.

My bad, the marker+linestyle+color string shortcut as 3rd arg is only available for plot, not for Line2D, it needs to use the keyword arguments only:

```
reward_artist = Line2D([], [], c="k", ls="none", marker="o", ms=s, mew=lw, label="Reward")
no_reward_artist = Line2D([], [], ls="none", marker="o", mfc="w", mec="k", ms=s, mew=lw, label="No reward")
Qvalue_artist = Line2D([], [], c="k", ls="-", lw=3, alpha=0.3, label="Qvalue (centered)")

ax.legend(handles=[no_reward_artist, Qvalue_artist, reward_artist], fontsize=12, loc=(1.01, 0.27));
```
> Also, I was not able to hide the plot code cell. AFAIK I need to use :tags: [hide-cell] in the myst.md file? but then the pre-commit complains because the notebook and the myst.md are not consistent ... can you please give me a tip here :) ?

One option is to edit the myst file (one of the reasons for having both myst and ipynb files is to be able to see and review such metadata, you can read https://github.com/pymc-devs/pymc-examples/pull/403#issuecomment-1193133893 if you are interested in this). But it is also possible to do from jupyter. Here you only have to do it once, so it isn't very critical which option you choose, if you had to update a lot of metadata I'd recommend editing the myst file. Overview of the options below:

* Editing myst. MyST and ipynb are synced, it is not that myst is purely an output of the jupyter notebook one. Therefore, if you have to update a lot of metadata in a notebook, I recommend making all the non-metadata changes from jupyter, then commiting that step. At this point, no format has any metadata so they are both perfectly synced, commit to make sure you don't lose anything. Then open the myst file and edit the metadata there, add the hidden cell tags, the figure captions and alt text... after that, run pre-commit which will sync the changes, this time modifying the ipynb file, and commit. The reason for the save+commit between the two is that if you edit both at the same time between syncing then the sync is broken and the only way to fix it is to lose the changes in one of the files, and also, pre-commit is tricky and runs only on staged changes, so if you run it and tries to sync the staged version it can edit the file and overwrite the unstaged one, which effectively means you lose the unstaged changes as these are the ones that are not tracked anywhere.
* Editing from jupyter. jupyter notebook, lab (and I assume vs code extension for jupyter and the like) have ways of editing the cell tags and metadata... This page explains it in more detail: https://jupyterbook.org/en/stable/content/metadata.html. However, the clicking for each cell and (in the case of metadata) having to write json instead of yaml is quite a pain in the ass, hence my recommendation to use myst if making significant metadata/tag changes.

myst_nbs/case_studies/reinforcement_learning.myst.md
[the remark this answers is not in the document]
ricardoV94: "Fitting two action... "? Otherwise it's okay to simply leave the number out. I don't like the "Fitting a two variable ... " title

myst_nbs/case_studies/reinforcement_learning.myst.md
juanitorduz: Thanks! I added it in https://github.com/pymc-devs/pymc-examples/pull/410/commits/5252c2b4d8304d20b7eb5daf2ccafdc60e229f0c (I am always hesitant to commit suggestions in jupyter notebooks 😄 )

myst_nbs/case_studies/reinforcement_learning.myst.md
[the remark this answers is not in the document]
juanitorduz: Thank you very much for such detailed explanation! (Is this somewhere in the contributing guidelines?) 🚀 I think https://github.com/pymc-devs/pymc-examples/pull/410/commits/cdaecc8003503bb50887670839ceb06299667d69 solved this!

Also, the plot suggestion worked perfectly!

juanitorduz:
Ok! I think the style is in good shape for a first review iteration in case someone wants to add or suggest more comments on the content. I have done no change (besides style)

**Remark:** Seems the title is a bit long? In the gallery id does not render nicely. Maybe is better if we remove the "two action" from the title?

ricardoV94:
> Remark: Seems the title is a bit long? In the gallery id does not render nicely. Maybe is better if we remove the "two action" from the title? 

Agree

myst_nbs/case_studies/reinforcement_learning.myst.md
```diff
@@ -0,0 +1,540 @@
...
+
+## References
+
+:::{bibliography}
+:filter: docname in docnames
+:::
+
++++
+
+## Credits
```
OriolAbril, line 533: All the credits/authorship info should be here, not divided in two different sections in the top/bottom of the notebook respectively

juanitorduz: Thanks! Fixed in https://github.com/pymc-devs/pymc-examples/pull/410/commits/49b3d6b503726f242dc529e7cb3d2a15c0d2bc55

OriolAbril: The section was not redundant and it should all be placed here. Some deviations are ok but notebooks should follow the jupyter style guide as closely as possible. This includes: https://docs.pymc.io/en/latest/contributing/jupyter_style.html#authorship-and-attribution

juanitorduz: Ups! Sorry, I was following one of the recent notebooks. I will bring it bak again 🙇

juanitorduz: Ok, here are all the credits together https://github.com/pymc-devs/pymc-examples/pull/410/commits/61e788123690b4b205c960c819a92b126dfa17e8

juanitorduz:
<p>Following suggestion I added some model comparison and a posterior predictive check for the Bernoulli model, see https://github.com/pymc-devs/pymc-examples/pull/410/commits/7d9dd5681a33096cae8dac05f587b2e5305c8c12</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/410/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

OriolAbril:
<p>I would recommend using https://python.arviz.org/en/latest/api/generated/arviz.plot_separation.html which is designed for binary outcomes.</p><p> </p><p>For completeness, here is what the plot above is showing. The black line is the histogram of the observations. As the dtype is integer, ArviZ will not use bins with width smaller than 1. There are therefore two bins: [0, 1) and [1, 2) that show it is more probable to observe a 1 than a 0. The blue lines represent the same thing, but for a specific chain+draw combination, if the model is explaining the data generating process, the black line should be undistinguishable from these blue ones. The orange one is the histogram but over all chains an draws instead of a single one which can sometimes be informative too.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/410/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
Thank for your comments! I misunderstood the suggestion for this section 🤦 (apologies).
I will remove these confusing plots and add @ricardoV94 's comments ad part of the last section so that we do not make this notebook unnecessary long. I think this could be a good first iteration 😅

juanitorduz:
<p>Better comment added in https://github.com/pymc-devs/pymc-examples/pull/410/commits/e9035c7611e59cd831e5a70255251dbb3f2f28d0</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/410/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>Agree! Thanks! But probably not with the UPPER CASE right? Seems a bit to aggressive XD. </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/410/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

juanitorduz:
<p>added in https://github.com/pymc-devs/pymc-examples/pull/410/commits/f72af486f44452e877e835caa5965a480f4affba</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/410/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94 (approved):
Approved content wise. I leave it to @OriolAbril to confirm the meta stuff is correct

examples/references.bib
```diff
@@ -48,6 +48,24 @@ @misc{carpenter2016hierarchical
...
+  author        = {Wilson, Robert C and Collins, Anne GE},
+  editor        = {Behrens, Timothy E},
+  volume        = 8,
+  year          = 2019,
+  month         = {nov},
+  pub_date      = {2019-11-26},
+  pages         = {e49547},
+  citation      = {eLife 2019;8:e49547},
+  doi           = {10.7554/eLife.49547},
+  url           = {https://doi.org/10.7554/eLife.49547},
```
OriolAbril, line 62: I think only one of them should be present, I'd keep the doi only. Both are rendered but provide the same info so it looks duplicated: https://pymcio--410.org.readthedocs.build/projects/examples/en/410/case_studies/reinforcement_learning.html#references

myst_nbs/case_studies/reinforcement_learning.myst.md
```diff
@@ -0,0 +1,548 @@
...
+
+* Authored by [Ricardo Vieira](https://github.com/ricardov94) in June 2022
+
+* Adapted PyMC code from Maria Eckstein ([GitHub](https://github.com/MariaEckstein/SLCN), [PyMC Discourse](https://discourse.pymc.io/t/modeling-reinforcement-learning-of-human-participant-using-pymc3/1735))
+
+* Adapted MLE code from Robert Wilson and Anne Collins {cite:p}`collinswilson2019` ([GitHub](https://github.com/AnneCollins/TenSimpleRulesModeling))
```
OriolAbril, line 541: I was thinking about making these two sub-bullet points of the first one, what do you thing?

OriolAbril (approved):
left a couple more nits, looks great, thanks

myst_nbs/case_studies/reinforcement_learning.myst.md
```diff
@@ -0,0 +1,548 @@
...
+---
+
++++ {"id": "Pq7u0kdRwDje"}
+
+(reinforcement_learning)=
+# Fitting a Reinforcement Learning Model to Behavioral Data with PyMC
+
+:::{post} Aug 5, 2022
+:tags: Aesara, Reinforcement Learning
+:category: advanced
```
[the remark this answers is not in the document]
juanitorduz: Added in https://github.com/pymc-devs/pymc-examples/pull/410/commits/053036a309e498461f25e0f6748c281334c1f706

examples/references.bib
juanitorduz: Removed in https://github.com/pymc-devs/pymc-examples/pull/410/commits/053036a309e498461f25e0f6748c281334c1f706

myst_nbs/case_studies/reinforcement_learning.myst.md
juanitorduz: Yes! Added in https://github.com/pymc-devs/pymc-examples/pull/410/commits/053036a309e498461f25e0f6748c281334c1f706

PR #412, opened 2022-08-07
Notebook: model_averaging

reshamas:
# {Insert Description}

### Helpful links


OriolAbril:
Not sure if I should comment here or on #411, I hope here is fine. The root of your problems is not pre-commit but git itself.

You can see that the PR is suggestiong changes for 4 files, not 2 only as it should. There are changes to the bayes factor notebook that get included in this PR, and in addition, the changes in that notebook are git conflict messages which break the ipynb file as they are not valid json. Therefore, most of the pre-commit tasks fail, not only jupytext, see [black error message](https://github.com/pymc-devs/pymc-examples/runs/7712961347?check_suite_focus=true#step:5:238) or [isort error message](https://github.com/pymc-devs/pymc-examples/runs/7712961347?check_suite_focus=true#step:5:247) and how they are 1) related to bayes factor notebook, not to model averaging one and 2) related to json errors, not to the check failing.

If I look at https://github.com/reshamas/pymc-examples/tree/main, I see a message that says "This branch is 5 commits ahead of pymc-devs:main.", which is a bad indicator. That means that all new branches you generate from main will contain those extra commits _even before you make the changes you want to submit a PR for_, and those changes are the git conflict json breaking ones. My recommendation is to delete the main branch in your local repository and create it again from the main branch in upstream (aka pymc-examples, not the one on your fork). Eventually, force push to origin (do not pull before force pushing or you'll be back at step 0) so that your fork also has its main branch synced with the main branch in pymc examples.

WARNING: I have not tested these commands, but I belive you'll need to do something like the commands below _while not being on the main branch_

```
git fetch upstream
git branch -D main
git checkout -b main upstream/main
# here git status should say your are now on branch main that tracks upstream/main
git branch main --set-upstream-to=origin/main
# here git status should say you are on branch main that tracks origin/main
git push --force
```

from here you should now be able to create new PRs without git conflicts or broken CI from the first commit. To fix this PR you would need to rebase (not exactly sure how though, would need to see it firsthand) or alternatively copy the model averaging notebook, create a new branch from main and then paste the notebook onto the new branch so you can open a new PR without changes to bayes factor, only to model averaging.

PR #424, opened 2022-09-24
new Difference In Differences notebook

drbenvincent:
This adds a new notebook covering the difference in differences approach to causal inference.


lucianopaz (changes requested):
It looks very nice @drbenvincent! I left two comments of things that could be clarified.
On the other hand, I would have loved if you had shown the results of an alternative simpler model. For example the one that only looks at the treatment group outcome, or some other naive method that doesn’t think about treatment time or neglects the pre test observations. Those should return biased or highly uncertain estimates of treatment effects, while the difference in difference should be unbiased and have low variance.
If you feel that doing that is out of scope of an example, it’s fine.

Armavica:
Thank you for the notebook which was an interesting read! I found a few typos and I think similarly to @lucianopaz I wonder about a simpler model which I think is equivalent to this method, but I may be missing something. I don't really understand the distinction between `group` and `treatment` in this particular example, but perhaps this distinction is useful for more complex data?

drbenvincent:
<p>Thanks. Fixed in an upcoming commit</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/424/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Hi</p><p>The DAG is the same as Fig 18.2 from The Effect <a href="https://theeffectbook.net/ch-DifferenceinDifference.html" target="_blank">https://theeffectbook.net/ch-DifferenceinDifference.html</a></p><p> </p><p>In an upcoming commit I've added explanation for why treatment and group are different. Namely, the treatment group is only treated after the intervention time.</p><p> </p><p>I think a 3-variable setup could match the results, but you're thinking statistically, not causally. There is a bit of a mindset shift involved. The 4-variable model captures the actual causal relationships. The idea is that if you intervene (with the do operator) then all the child nodes should be updated accordingly.</p><p> </p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/424/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>resolved in an upcoming commit</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/424/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Thanks. Fixed in an upcoming commit</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/424/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Did a bit of rearranging with the time variables</p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/424/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Changed to "disambiguate"</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/424/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
Thanks for the feedback @lucianopaz and @Armavica. I think I've addressed all the typos and points of confusion.

In terms of a simpler model, I think I might come back and look at that another time. My focus at the moment is (when I have time for these notebooks) to try to lay down a number relating to causal inference. And then loop back around at some point when I've got more feedback and my understanding has increased.

Armavica:
<p>Thank you for your answer. I think that my main confusion is the inclusion of "time" as an explicit variable of this DAG, but it might only be a concern with the DiD method and not with your work. From several sources (<a href="https://en.wikipedia.org/wiki/Directed_acyclic_graph#Causal_structures" target="_blank">the DAG wikipedia article</a>, the Statistical Rethinking book, etc.) I understood that the nodes of a causal model describe events taking place at specific times, and that arrows mean "... has a direct causal influence on ...". Causality ensuring that arrows are only drawn from earlier events/nodes to later events/nodes, the graph is acyclic by construction. Time is an intrinsic component of the model because it imposes the direction of the arrows. But including an explicit node "time" breaks this framework and I don't understand what this kind of DAG means anymore.</p><p> </p><p>Indeed, what should "time has a direct causal influence on ..." mean? And what about the opposite: could one ever draw an arrow towards the "time" node, or is the "time" node special in the sense that no arrow can point to it? And what ensures that the rest of the graph is acyclic, if arrows do not have this intrinsic constraint of going from an earlier event to a later event?</p><p> </p><blockquote>I think a 3-variable setup could match the results, but you're thinking statistically, not causally. There is a bit of a mindset shift involved. The 4-variable model captures the actual causal relationships. The idea is that if you intervene (with the do operator) then all the child nodes should be updated accordingly.</blockquote><p> </p><p>This could very well be the case: I am also learning. But I fail to understand why this 3-variable model would not capture the causal relationships, or would not allow to update child nodes after an intervention. Changing the initial metric has causal repercussions on the treatment decision as well as on the final value, and changing the treatment variable has causal repercussions on the final value. All of the arrows in this model are causal because of the direction of time.</p><p>However, in the 4-variable model with the "time" node, because the initial metric and final metric are represented with the same variable "mu" (node "outcome"), how would one describe an intervention on the initial metric with the do operator? Also, is it really causal to have "treatment" and "group" point at "outcome" if "outcome" can represent both the initial and the final value, even though "treatment" and "group" cannot causally influence the initial value?</p><p> </p><p>I apologize if my questions don't make sense or if they are trivial, and I also understand that this notebook is not meant to be a causal inference textbook, so please feel free to ignore my comments if they are too "philosophical" or irrelevant!</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/424/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Armavica:
It looks like you define a `RANDOM_SEED` and a seeded prng in the beginning of the notebook but never use them later, or am I missing something?

drbenvincent:
> It looks like you define a `RANDOM_SEED` and a seeded prng in the beginning of the notebook but never use them later, or am I missing something?

Thanks, I forgot the style guide... I should be using `rng.normal` which uses the random seed, rather than `scipy.stats.normal`.

drbenvincent:
No, these are good questions and I am certainly not at expert level yet. And I may have been a bit hasty with my reply about statistical vs causal thinking. (Although that might be an interesting thing to write about separately)

Your concerns seem to be focussed on 'unfolding nodes over time' vs not doing that. I think there are different conventions about this. Maybe it makes sense if you just consider two discrete points in time (pre/post) but it's less workable if you want to evaluate over continuous (or many discrete values of) time.

I don't know if I can yet fully articulate an decent answer at this point in my learning, but a few points:
- The way in which the time node is thought about is expanded on in this video https://www.youtube.com/watch?v=ggYnrOGG97o
- I think we can separate different ideas here. As you say, an edge can be thought of as representing time because of temporal precedence. But that seems like a slightly different notion of something changing as a function of time. It doesn't seem that tricky if you just think $time \rightarrow treatment$ as describing $treatment = f(time)$ for example.
- Yes, it's a fair point to say that the post intervention outcome can be causally influenced by the pre intervention outcome and that it would make sense to have represent this as outcomes unfolded over time. But like I say, I think this is a convention thing and different approaches might be better suited in different situations.

I think that's all I've got at this point. Onwards with the learning. Like I say, I hope to loop back around my notebooks in the future to disambiguate and further clarify. But yes, have to not fall into the trap of trying to write a textbook chapter!

Armavica:
Yes, perhaps it is just a convention, in which case whatever seems more intuitive for every particular modeler and model should work :) Thank you for this discussion in any case, it was interesting to learn about this framework!

twiecki:
Just read through it and could follow everything nicely. One thought I had was that instead of just a linear model it could be e.g. a GP to model a more complex time-series. Of course that's outside the scope of this post, and I'm not sure if it should even be mentioned.

drbenvincent:
> Just read through it and could follow everything nicely. One thought I had was that instead of just a linear model it could be e.g. a GP to model a more complex time-series. Of course that's outside the scope of this post, and I'm not sure if it should even be mentioned.

Thanks.

I think GP's could be more relevant in an interrupted time series type design perhaps. So I think I'll not mention it here

drbenvincent:
<blockquote>I’m a bit fuzzy about the DiD model. If you had in fact been able to randomise the group membership, what would have changed in the model?&nbsp;</blockquote><p>If there was randomisation then we are outside of the realm of quasi-experiments. So you could do something more like A/B tests in that case.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/424/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<blockquote>Am I correct in thinking that the DiD only assumes that the intercept changes? The slope is assumed to be shared, and then there’s the effect size, so the only thing left is the intercept. This sounds like I’m missing something else. </blockquote><p>The parallel trends assumption assumes that both groups are evolving identically _in the absence of treatment_. But in the presence of treatment there is some unknown deflection of the treatment group, so there is a free parameter for the slope of the treatment group.</p><p>Then you also have a free parameter for an intercept difference. This can be interpreted as capturing differences in the groups (prior to treatment) which might be expected because of the lack of randomisation.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/424/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<blockquote>Could it be that DiD also gives unbiased estimates of treatment effects even if you have unbalanced recordings between the groups for the different times (E.g more control measurements in pre, more treatment measurements in post)?</blockquote><p>As far as I understand it, the assumption is that you have panel / repeated measures data. Ie you have data points for each unit at both pre and post treatment times. But I think there's nothing in the model formulation at the moment which relies upon this. At the moment you <em>could</em> have unbalanced recordings between the groups for the different times.</p><p> </p><p>But that does actually seem odd to me now that you bring it up. Wouldn't it be more sane to have a hierarchical model where you are modelling the slopes and intercepts at the unit and the population level and you draw your conclusions from the population level parameters? What do you think? Maybe that's worth a note and perhaps a different follow up hierarchical difference in differences example notebook?</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/424/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
> View / edit / reply to [this conversation](https://app.reviewnb.com/pymc-devs/pymc-examples/pull/424/discussion/) on ReviewNB
> 
> _lucianopaz commented on 2022-09-29T20:05:25Z_ ----------------------------------------------------------------
> 
> I thought that you were going to use the minimum wage and unemployment dataset here. I’m confused, why did you have to talk about that dataset if you weren’t going to use it?

I just wanted to give a tangible example in case some found the rest of the introduction too abstract. But I could always add another one or two examples if you thought that having the one sets up the wrong expectations?

twiecki (approved):

PR #457, opened 2022-11-03
update BART example

aloctavodia:
This mainly reflects changes in the new API.


myst_nbs/case_studies/BART_introduction.myst.md
```diff
@@ -33,8 +33,8 @@ print(f"Running on PyMC v{pm.__version__}")
 ```
 
 ```{code-cell} ipython3
-RANDOM_SEED = 8457
-rng = np.random.RandomState(RANDOM_SEED)
+RANDOM_SEED = 5781
+np.random.seed(RANDOM_SEED)
```
OriolAbril, line 37: I think the global seed doesn't have any effect on v4? if the rng is not used then we can only define the random seed variable to pass to `pymc.sample`

aloctavodia: I double-checked and it has an effect on BART.

ricardoV94: It shouldn't ;)

aloctavodia: It is not clear to me what I have to change on pymc-bart, but I will give it a try. Any suggestion is welcome.

OriolAbril (approved):

PR #459, opened 2022-11-05
Correct definition of nu and assign lower value for sigma_low

hyosubkim:
- 1 should be added to nu_minus_one using pm.Deterministic, otherwise addition of 1 in old definition has no effect and hence probability was non-zero for values of nu that are less than one
- Added definition and figure panel for log10(nu) to better align with Kruschke 2013
- Lowered sigma_low value to 10^-1 because sigma values less than 1 are credible (as seen for Group 2 Std Dev in Fig 3 from Kruschke 2013)

With suggested changes, results will more closely align with Kruschke paper, which example is based on.

# {Insert Description}

### Helpful links


OriolAbril:
Thanks, can you run pre-commit?

It should be something like:

```
pip install pre-commit # or conda or local install
pre-commit run --all
```

which should update the `.myst.md` file. Then commit and push and all checks should pass hopefully

hyosubkim:
I did as suggested and pushed the changes. Hope it went through okay this time.

OriolAbril (approved):

OriolAbril:
Thanks! This got both CI jobs to pass

hyosubkim:
Great! Thanks again for your help

PR #463, opened 2022-11-12
Multilevel revision

fonnesbeck:
This is an update to the multilevel modeling notebook that makes two main changes:

1. Removes the independent floor vs basement parameterization from the current notebook. Expressing the floor and basement measurements as entirely independent is not a great starting point, given that they are measuring the exact same phenomenon, just from a different location. This reverts to the covariate model that simply uses floor as an offset. I think this is less confusing overall.
2. Adds the non-centered parameterization. While perhaps redundant with another notebook, its an important component of modeling this problem correctly. 

I am also removing the `GLM-hierarchical` notebook, as it is almost completely redundant with this one, now that it no longer uses the extinct `glm` submodule.


fonnesbeck:
Some bad merge artifacts. Will create new PR from clean branch.

PR #469, opened 2022-11-21
Remove normalizing flows notebook

fonnesbeck:
As normalizing flows were removed for v4, this PR removes the notebook that demonstrates how normalizing flows are used.


OriolAbril:
looks good, but the myst paired notebook should be removed too

PR #373, opened 2022-06-09
Rerun vi

ferrine:
related to #333

PR #470, opened 2022-11-23
Updates to VI quickstart

fonnesbeck:
Adds some contextual material and a simple example (rather than jumping straight into mixtures).


OriolAbril:
would you consider taking over #373 or do you prefer going through the vi notebooks one by one?

PR #473, opened 2022-12-07
some viz changes to multilevel notebook

OriolAbril:
Some follow-up changes after #468, it is an amalgam of minor changes like:

* Add the target from the glm-hierarchical  notebook that was removed here. This makes other
  notebooks that referenced it link here now.
* Use bibtex references throughout the text
* Use arviz labellers in plot_forest for better looking plots

I have also deleted the myst pairing of the glm-hierarchical notebook which was still there.


fonnesbeck (approved):
Looks great!

Issue #407, opened 2022-08-01
Bayesian Decision Analysis (Posterior Predictive + Utility Function)


LarsKarbach:
# Notebook proposal
**Title**: Bayesian Decision Analysis

## Why should this notebook be added to pymc-examples?
PyMC lacking an example of a Bayesian Decision Analysis, much in the style of an example in Stan's Users Guide (see reference). I could try to create a Notebook, if that's of interest.

# References
https://mc-stan.org/docs/stan-users-guide/example-decision-analysis.html

canyon289:
this would be great! Please go ahead and add one

drbenvincent:
Seconded, having an example on this topic would be fantastic. Is this still on your radar @LarsKarbach?

LarsKarbach:
Yes, altough i let it slip a bit. It's time to change this! How about give me time until the end of November and see from there if i need help?

OriolAbril:
Took a stab at this in #477, still need to flesh out the text but reviews will be very welcome at any stage.

LarsKarbach:
Hi, very cool. I currently work on an example, which is not running yet. I will continue with this. If we choose to not go with mine, its not to bad.

PR #477, opened 2022-12-15
add decision utility funciton draft

OriolAbril:
+ [x] PR description contains a link to the relevant issue: taking a stab at #407


drbenvincent:
* Looks great. I guess there will be some more explanation text added in the final draft?
* There's potentially scope for doing a simple model 1 with just one outcome variable as a _really_ simple start, then bring in the model with 3 outcome variables to demonstrate the generality.
* Could be worth describing where the utility function comes from, or that this is up to what the person finds most important.

OriolAbril:
> I guess there will be some more explanation text added in the final draft?

Yes, this is only getting the code to work and sharing in case someone had suggestions on the model. I'll try to come back to this and add the explanation and finish t some point next week

fonnesbeck:
Dead or alive?

OriolAbril:
Probably best to consider it dead. I do still have it on the TODO but with many higher priority things before it, I don't think I'll get to it anytime soon. It is also ok with me if someone wants to take over and finish it too.

PR #481, opened 2022-12-24
Rename aesara to pytensor.

twiecki:
This used the following script:
```bash
#!/bin/bash

if [ -z "$1" ]
  then
    echo "Script to automatically convert code (*.py and *.ipynb) from PyMC3 to 4.0. Use with care."
    echo "Usage: pymc3_to_4.sh <path>"
    exit 1
fi

declare -a replace_strings=(
    "s/from aesara import tensor as at/import pytensor.tensor as at/g"
    "s/import aesara\.tensor as at/import pytensor.tensor as at/g"
    # "s/tt\./at./g"
    "s/aesara/pytensor/g"
    "s/Aesara/PyTensor/g"
    "s/https\:\/\/github.com\/aesara-devs\/aesara/https\:\/\/github.com\/pymc-devs\/pytensor/g"
)

for replace in "${replace_strings[@]}"; do
    find $1 -name "*.ipynb" -type f -exec sed -i -e "/png/!$replace" {} \;
    find $1 -name "*.py" -type f -exec sed -i -e "/png/!$replace" {} \;
done
```

Then ended up in myst-hell, deleted the myst-dir, and re-committed. Some myst-files were deleted, are those obsolete? Anyway I can tell jupytext to just overwrite whatever is in the myst file? This process is so annoying and confusing.

I also didn't change any `at.` references, these are a bit tricker and I'd like to do those separately.


OriolAbril (approved):
We should still aim to rerun the notebooks, the watermark will be broken now for these notebooks

Issue #443, opened 2022-10-12
Add template NB


twiecki:
There are quite a few things that have to be present in a NB, we should have a template NB that has all of them.

drbenvincent:
Would presumably have to be stored in a location such that it's not treated as an actual notebook and thus rendered on the site.

PR #487, opened 2022-12-28
add template notebook

drbenvincent:
Addresses #443. After this is merged, we should mention it in the Style Guide. I've made this about as minimal as possible just to get things started. Happy to take on board any suggestions.

### DIFF
--- a/template_notebook.ipynb
+++ b/template_notebook.ipynb
@@ -153,7 +153,8 @@
    "id": "1e4386fc-4de9-4535-a160-d929315633ef",
    "metadata": {},
    "source": [
-    ":::{include} ../page_footer.md :::"
+    ":::{include} ../page_footer.md",
+    ":::"
    ]
   }
  ],
--- a/template_notebook.ipynb
+++ b/template_notebook.ipynb
@@ -1,0 +1,186 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "domestic-remove",
+   "metadata": {},
+   "source": [
+    "(template_notebook)=\n",
+    "# This is a template notebook\n",
+    "\n",
+    ":::{post} January, 2023\n",
+    ":tags: binomial regression, generalized linear model, \n",
+    ":category: beginner\n",
+    ":author: Jane Doe\n",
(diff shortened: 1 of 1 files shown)


template_notebook.ipynb
```diff
@@ -0,0 +1,186 @@
...
+   "cell_type": "markdown",
+   "id": "domestic-remove",
+   "metadata": {},
+   "source": [
+    "(template_notebook)=\n",
+    "# This is a template notebook\n",
+    "\n",
+    ":::{post} January, 2023\n",
+    ":tags: binomial regression, generalized linear model, \n",
+    ":category: beginner\n",
```
OriolAbril, line 14:
```suggestion
    ":category: beginner, reference\n",
```

template_notebook.ipynb
```diff
@@ -0,0 +1,186 @@
...
+    "%load_ext watermark\n",
+    "%watermark -n -u -v -iv -w -p pytensor"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1e4386fc-4de9-4535-a160-d929315633ef",
+   "metadata": {},
+   "source": [
+    ":::{include} ../page_footer.md :::"
```
OriolAbril, line 156:
```suggestion
    ":::{include} ../page_footer.md",
    ":::"
```

template_notebook.ipynb
```diff
@@ -0,0 +1,186 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "level-balance",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "%config InlineBackend.figure_format = 'retina'\n",
```
OriolAbril, line 53: I am not sure this is needed but I don't know what it's supposed to do either

drbenvincent: It's not needed, but it sets the plot resolution to be high, appropriate for those with high res displays (eg on macs). I'll add a comment clarifying this, but would be ok with removing it. But preference is to keep... standard low res stuff looks bad on Apple monitors.

OriolAbril (approved):

PR #490, opened 2023-01-04
notebook: Bayes Factor; run using pymc v5

reshamas:
# Description
Run Bayes Factor notebook with PyMC version 5.

NOTE:  needed to set `progressbar=False` to avoid errors.


cluhmann:
Is there an issue open for the progressbar problem?

reshamas:
> Is there an issue open for the progressbar problem?

No. @OriolAbril said it was specific to this notebook, not a general problem.

OriolAbril:
There is an issue: https://github.com/pymc-devs/pymc/issues/5855 we already commented there, but it is not clear how to reproduce the issue so it hasn't been fixed.

cluhmann:
Got it.  Just wanted to make sure it was being referenced here if the issue already existed.  But now it is!

OriolAbril:
I have re-executed the notebook locally to see if I could reproduce the issue, but I didn't, in my case it worked, so the issue will continue to be open and with no solution in sight. I have pushed so we can merge without needing to remove the progressbar in sample_smc

OriolAbril (approved):

Issue #474, opened 2022-12-14
Bayesian Methods for Reliability Data


NathanielF:
# Notebook proposal
**Bayesian Methods fo reliability Data**

This topic crops up in engineering quite a bit and generally involves a spin on survival analysis  (time to failure) type models, but it's also tightly linked with the notion of cost-benefit, since there is a tolerable degree of failure. I was thinking about riffing on a discussion about prediction intervals for failure analysis in:

https://www.wiley.com/en-us/Statistical+Methods+for+Reliability+Data,+2nd+Edition-p-9781118115459

and contrast the frequentist and Bayesian approach to calibrated prediction. It might be interesting to contrast conformal prediction methods too, since it is a frequentist approach to uncertainty quantification which is getting quite a bit of coverage lately. 

Suggested categories:
* Level: Intermediate
* [Diataxis](https://diataxis.fr/) type: explanation

## Related notebooks
It might have cross over in themes with: https://github.com/pymc-devs/pymc-examples/issues/407
and some of the techniques for modelling the failure predictions draw on survival analysis models.

OriolAbril:
I think it would be a great addition, and it is a great fit for an explanation type notebook (that can/could complement other how-to notebooks on generating predictions for different kind of models for example).

NathanielF:
Did a little work on this over the Christmas holidays. Written up a brief discussion of nonparametric and parametric estimation of reliability distributions and their CDFs. I'm following an example in the book cited above which moves through the frequentist MLE style estimation to a data set with very few failures... it argues for the importance of Bayesian modelling in this case especially.... so I think it's a good example for here.

However, I'm stuck a bit on trying to replicate the model fits achieved in the book or bring them anywhere close to the MLE fits. I'm using a Weibull survival model and I've tried (a) to replicate the stan model they use in the book (b) just use a base pymc  weibull fit adding a potential for lcdf portion of the log likelihood and (c) use the censored model transformation as discussed in one of the survival analysis notebooks by @drbenvincent . I've use both period form and item-period data sets as described in the notebook, but none seem to recover good model fits with the pymc implementation... 

I reckon I must be doing something silly so would appreciate any pointers you might have.

NathanielF:
Think the above discussed issue is resolved now: 
<img width="699" alt="image" src="https://user-images.githubusercontent.com/10588167/210185102-fd20a805-4451-475b-af5e-47e2d62e35a1.png">

[...]

PR #491, opened 2023-01-06
Reliability and Calibrated Prediction

NathanielF:
This will be an example notebook demonstrating the techniques of survival analysis used in reliability context with a focus on the predictive distribution and calibrated prediction intervals from a Bayesian and MLE perspective. Relates to issue: https://github.com/pymc-devs/pymc-examples/issues/474

Adding this draft version here, because i'm having a little trouble with the pre-commit checks

+ [X] PR description contains a link to the relevant issue:

### DIFF
--- a/examples/case_studies/reliability_and_calibrated_prediction.myst.md
+++ b/examples/case_studies/reliability_and_calibrated_prediction.myst.md
@@ -1108,7 +1108,7 @@ output_df
 ```{code-cell} ipython3
 def cost_func(failures, power):
     ### Imagined cost function for failing item e.g. refunds required
-    return -np.power(failures, power)
+    return np.power(failures, power)
 
 
 mosaic = """AAAA
@@ -1176,6 +1176,7 @@ ax2.hist(
     bins=20,
 )
 ax2.set_xlabel("$ cost")
+# ax2.set_xlim(-60, 0)
 ax2.legend()
 ```
 
--- a/examples/case_studies/reliability_and_calibrated_prediction.myst.md
+++ b/examples/case_studies/reliability_and_calibrated_prediction.myst.md
@@ -1082,7 +1082,7 @@ print("Upper Bound 95% PI:", binom(1700, rho).ppf(0.95))
 
 ### Applying the Same Procedure on the Bayesian Posterior
 
-We'll use the posterior predictive distribution of the uniformative model. We show here how to derive the uncertainty in the estimates of the 95% prediction interval for the number of failures in a time interval.
+We'll use the posterior predictive distribution of the uniformative model. We show here how to derive the uncertainty in the estimates of the 95% prediction interval for the number of failures in a time interval. As we saw above the MLE alternative to this procedure is to generate a predictive distribution from bootstrap sampling. The bootstrap procedure tends to agree with the plug-in procedure using the MLE estimates and lacks the flexibility of specifying prior information.
 
 ```{code-cell} ipython3
 def PI_failures(joint_draws, lp, up, n_at_risk):
--- a/examples/case_studies/reliability_and_calibrated_prediction.myst.md
+++ b/examples/case_studies/reliability_and_calibrated_prediction.myst.md
@@ -5,9 +5,9 @@ jupytext:
     format_name: myst
     format_version: 0.13
 kernelspec:
-  display_name: Python [conda env:pymc_ar_ex] *
+  display_name: Python 3 (ipykernel)
   language: python
-  name: conda-env-pymc_ar_ex-py
+  name: python3
 substitutions:
   extra_dependencies: lifelines
 ---
@@ -19,6 +19,9 @@ substitutions:
 :tags: time-to-failure, prediction, calibration, survival analysis, censored
 :category: intermediate
 :author: Nathaniel Forde
+:::
+
+:::{include} ../extra_installs.md
 :::
 
 ```{code-cell} ipython3
@@ -358,9 +361,9 @@ def plot_cdfs(actuarial_table, dist_fits=True, ax=None, title="", xy=(3000, 0.5)
     )
     ax.plot(actuarial_table["t"], actuarial_table["CI_95_ub"], color="darkorchid", linestyle="--")
     ax.fill_between(
-        actuarial_table["t"],
-        actuarial_table["CI_95_lb"],
-        actuarial_table["CI_95_ub"],
+        actuarial_table["t"].values,
+        actuarial_table["CI_95_lb"].values,
+        actuarial_table["CI_95_ub"].values,
         color="darkorchid",
         alpha=0.2,
     )
@@ -375,9 +378,9 @@ def plot_cdfs(actuarial_table, dist_fits=True, ax=None, title="", xy=(3000, 0.5)
         actuarial_table["t"], actuarial_table["logit_CI_95_ub"], color="royalblue", linestyle="--"
     )
     ax.fill_between(
-        actuarial_table["t"],
-        actuarial_table["logit_CI_95_lb"],
-        actuarial_table["logit_CI_95_ub"],
+        actuarial_table["t"].values,
+        actuarial_table["logit_CI_95_lb"].values,
+        actuarial_table["logit_CI_95_ub"].values,
         color="royalblue",
         alpha=0.2,
     )
@@ -401,7 +404,7 @@ def plot_cdfs(actuarial_table, dist_fits=True, ax=None, title="", xy=(3000, 0.5)
 ```
 
 ```{code-cell} ipython3
-plot_cdfs(actuarial_table_shock, title="Shock Absorber")
+plot_cdfs(actuarial_table_shock, title="Shock Absorber");
 ```
 
 This shows a good fit to the data and implies, as you might expect, that the failing fraction of shock absorbers increases with age as they wear out. But how do we quantify the prediction given an estimated model?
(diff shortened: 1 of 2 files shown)


NathanielF:
Hi @drbenvincent and @OriolAbril 

I think this is ready for review now. It's longer than i anticipated and i've cut out the idea to compare against conformal prediction as there is already quite a bit of detail in there.

The broader structure has three parts: 

image

In the first part i'm just introducing time-to-failure data and the descriptive statistical analogues of survival and cdf functions. I follow a trajectory of first showing how we can analyse this kind of data with MLE or bootstrap style inference and then show on a trickier example with sparser data why the ability to add Bayesian priors to help calibrate the risk the profile is useful. Then i conclude with a "lets all be friends" kind of ending. 

I'm not sure who is best placed to review, but since i deal with censored data and I know @drbenvincent recently updated the censored data notebook, i thought you might find it interesting? Curious about what you think too @OriolAbril? Happy to take any feedback on the content or structure...

examples/case_studies/reliability_and_calibrated_prediction.myst.md
```diff
@@ -0,0 +1,1191 @@
...
+print("Lower Bound 95% PI :", binom(1700, rho).ppf(0.05))
+print("Upper Bound 95% PI:", binom(1700, rho).ppf(0.95))
+```
+
+### Applying the Same Procedure on the Bayesian Posterior
+
+We'll use the posterior predictive distribution of the uniformative model. We show here how to derive the uncertainty in the estimates of the 95% prediction interval for the number of failures in a time interval.
+
+```{code-cell} ipython3
+def PI_failures(joint_draws, lp, up, n_at_risk):
```
OriolAbril, line 1065: I'll update the code to use xarray-einstats, the inputs are xarray dataset and scalars IIUC, so there should be no need for looping in order to use scipy distributions.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
```diff
@@ -0,0 +1,1191 @@
...
+    ec="black",
+    bins=20,
+)
+ax2.set_xlabel("$ cost")
+ax2.legend()
+```
+
+The choice of model in such cases is crucial. The decision about which failure profile is apt has to be informed by a subject matter expert because extrapolation from such sparse data is always risky. An understanding of the uncertainty is crucial if real costs attach to the failures and the subject matter expert is usually better placed to tell if you 2 or 7 failures can be plausibly expected within 600 hours of service. 
+
+# Conclusion
```
OriolAbril, line 1161: I think this should be level 2 header, otherwise the table of contents looks a bit weird: https://pymcio--491.org.readthedocs.build/projects/examples/en/491/case_studies/reliability_and_calibrated_prediction.html.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
```diff
@@ -0,0 +1,1191 @@
...
+
+In particular we've seen how the MLE fits to our bearings data provide a decent first guess approach to establishing priors in the Bayesian analysis. We've also seen how subject matter expertise can be elicited by deriving key quantities from the implied models and subjecting these implications to scrutiny. The choice of Bayesian prediction interval is calibrated to our priors expectations, and where we have none - we can supply vague or non-informative priors. The implications of these priors can again be checked and analysed against an appropriate cost function.
+
++++
+
+## Authors
+
+Nathaniel Forde
```
OriolAbril, line 1171: This should follow the same format as other notebooks: https://docs.pymc.io/en/latest/contributing/jupyter_style.html#authorship-and-attribution

OriolAbril (commented):
I have commented on the myst file, but the changes should be done in ipynb and then use pre-commit to update myst one.

General comment: I would use https://myst-nb.readthedocs.io/en/latest/render/hiding.html to hide the source of some of the code. i.e. the stringIO ones can be hidden, we would only need to split them so the `df_name` to show the rendered table is in its own cell right below (not hidden so readers can follow the notebook and get the variable name even without expanding the cell). Some of the plots I think could also have their source hidden.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
NathanielF: On this one... i think i kind of disagree. While it's not the most efficient way to do this operation, i wanted the parallel with the bootstrapping looping procedures to be clear. We're instantiating in a loop a reasonable model fit to the data in each iteration of the loop, and seeing the variation induced in downstream statistics... i think pedagogically this is simpler than hiding the similarity of the procedure in more efficient code....

examples/case_studies/reliability_and_calibrated_prediction.myst.md
NathanielF: Done, changed this.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
NathanielF: Updated to better reflect the format.

NathanielF:
Thanks so much for the review. @OriolAbril. I've quibbled with one of the requests but i think i've addressed the rest. I've added hide-input tags to a bunch of the data loads and some of the more involved plotting functions.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
NathanielF: Oops I think I misread this comment yesterday, did you want to try and streamline the loop? My only point was that I'd like to keep the analogy with the bootstrap procedure clear....

But I think I could just add more text explaining that too.

NathanielF: Ok, i've added some more text to highlight the relationship between the bootstrap and posterior predictive distributions if you want to work some xarray magic, that'd be cool! Thanks!

examples/case_studies/reliability_and_calibrated_prediction.myst.md
```diff
@@ -358,9 +361,9 @@ def plot_cdfs(actuarial_table, dist_fits=True, ax=None, title="", xy=(3000, 0.5)
     )
     ax.plot(actuarial_table["t"], actuarial_table["CI_95_ub"], color="darkorchid", linestyle="--")
     ax.fill_between(
-        actuarial_table["t"],
-        actuarial_table["CI_95_lb"],
-        actuarial_table["CI_95_ub"],
+        actuarial_table["t"].values,
```
OriolAbril, line 364: I had an error in matplotlib trying to run this, tried updating matplotlib and numpy and it didn't seem to fix the issue, so I added this values to continue ahead. I don't think it is a big issue though.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
```diff
@@ -21,6 +21,9 @@ substitutions:
 :author: Nathaniel Forde
 :::
 
+:::{include} ../extra_installs.md
```
OriolAbril, line 24: I noticed you had set the substitutions metadata but this include was missing. It should now add some auto instructions on installing lifelines.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
```diff
@@ -444,23 +447,6 @@ def bayes_boot(df, lb, ub, seed=100):
 ```
 
 ```{code-cell} ipython3
-def bootstrap(df, lb, ub, seed=100):
```
OriolAbril: This is the same function as above, and this one wasn't being used, so I deleted it.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
```diff
@@ -953,15 +939,11 @@ az.summary(idata_informative)
 ```
 
 ```{code-cell} ipython3
-joint_draws = az.extract_dataset(idata, group="posterior", num_samples=1000)[
-    ["alpha", "beta"]
-].to_dataframe()
+joint_draws = az.extract(idata, num_samples=1000)[["alpha", "beta"]]
```
OriolAbril, line 942: The fact that was a dataframe wasn't being used anywhere, so I changed that because now below we take advantage of the fact `joint_draws` is an xarray dataset.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
OriolAbril: Updated, and found a bug in xarray-einstats while doing it 😄 (I have fixed it already)

OriolAbril (commented):
I reran the notebook locally to update that sampling snippet. And did a couple extra changes that I have commented here in the review. Let me know what you think, they are all quite minor except for the sampling one.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
NathanielF: Oh, cool. Yeah, i forgot this,

examples/case_studies/reliability_and_calibrated_prediction.myst.md
NathanielF: Yeah, i think that's fine.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
NathanielF: That's fair.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
NathanielF: It was a slightly different function. One implemented the naive bootstrap and the other implemented the bayesian bootstrap with a flat dirichelet prior. But yeah, basically the same, so i'm fine getting rid of it.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
NathanielF: Is it this one: 

image

I updated to my xarray einstats version, but still getting an error when i try to re-run your new function?

NathanielF:
Thanks @OriolAbril I'm fine with all the changes you made, i understand the logic of the xarray-einstats version too, but i just couldn't get it to work on my machine and i'm unsure why? I've attached a screenshot above.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
NathanielF: Can recreate with just this call: 
image

NathanielF:
Just fixing a typo at the beginning. Still have same xarray issue.

examples/case_studies/reliability_and_calibrated_prediction.myst.md
NathanielF: Oh!!!! Sorry, i see the issue, the latest change is not on pip!!!! Sorry, slow this morning. 

image

Ehm, that's cool. I'm happy with the changes and you can merge if you want. But for reproducibility should we be concerned that the user can't easily install the required packages?

NathanielF: OK, cool! Let's do it. Thanks again for all your help on this one! I'm happy with it now if you are.

OriolAbril (approved):

NathanielF:
Woop, woop! Thanks 😊!

PR #507, opened 2023-01-25
Out-of-sample predictions for BART intro notebook

juanitorduz:
# Out-of-sample predictions for BART intro notebook

Closes https://github.com/pymc-devs/pymc-bart/issues/48

- Add a section regarding out-of-sample predictions for the existing `bikes` example.
- Add a comment regarding extrapolating with BART models.


aloctavodia:
Thanks! 

What about adding a first OOS example where you treat the problem as a regression and then a second one (the one you already wrote) showing that BART is not good at extrapolation?

aloctavodia:
I would not change what you already added. What I would add is an example of out-of-sample predictions using the bike dataset, with the only difference being that we treat it as a regression problem, not a time-series problem.

We can keep the coal mining example unmodified.

juanitorduz:
I added a regression-like example in https://github.com/pymc-devs/pymc-examples/pull/507/commits/ae24debc47d48e9b0c8de604d8eb7055a6e63b99

aloctavodia (approved):
This is fantastic! Thanks!

OriolAbril:
@juanitorduz just for completeness, you have added significant piece of new code, so you should add yourself to both the authors section at the bottom (like you did) and at the `:author:` metadata at the beginning. Feel free to update whenever you work on another notebook

Issue #530, opened 2023-03-16
ModelBuilder notebook is missing predicted regression lines


fonnesbeck:
**Using ModelBuilder class for deploying PyMC models**:
**https://www.pymc.io/projects/examples/en/latest/howto/model_builder.html**:

## Issue description
The last plot of the notebook, which is supposed to be predicted regression lines, currently only shows the data points:

![image](https://user-images.githubusercontent.com/81476/225667575-acca476a-8d16-4584-be4b-a807ab6f3690.png)

reshamas:
@BerylKanali is working on this with @Monsurat-Onabajo

PR #538, opened 2023-03-31
Update ModelBuilder to show regression lines

BerylKanali:
@cluhmann Here is the PR from yesterday's session. I worked on it with @Monsurat-Onabajo

Towards https://github.com/pymc-devs/pymc-examples/issues/530
cc: @fonnesbeck @reshamas


examples/howto/model_builder.myst.md
```diff
@@ -271,8 +271,12 @@ After using the `predict()`, we can plot our data and see graphically how satisf
 fig, ax = plt.subplots(figsize=(7, 7))
 ax.plot(
     x_pred,
+    marker="x",
+    label="predict",
+),
+ax.plot(
```
OriolAbril, line 277: The changes are modifying an x vs y plot to two line plots where the x axis is always an arange and the y axis is mapped to the values in the 1st plot and to the y values in the 2nd plot. Mapping these two different quantities to the same axis is definitely wrong and should be changed. I think what @fonnesbeck wanted in the issue was having both scatter x vs y of the predictions along with the x vs y of the regressions lines from the model that generated those predictions. It would have a similar look to the last image in https://www.pymc.io/projects/examples/en/latest/generalized_linear_models/GLM-robust.html but with the x_pred / pred_means["y"] as the scatter instead of truth valued and no true regression line

Monsurat-Onabajo: Ok, I get the picture now, I will work on it

twiecki:
We also recently changed the API of ModelBuilder (slightly), are you interested in bringing this example up-to-date? Here is the PR: https://github.com/pymc-devs/pymc-experimental/pull/131

Monsurat-Onabajo:
When I tried to plot the graph of the truth values and the predicted values in one figure, I got this graph and upon investing further, I noticed the truth value data is more of a wavelength dataset than a regression dataset. That was probably the reason a line can't be used as a connector because there is no pattern in the data for the graph to be formed on

twiecki:
@Monsurat-Onabajo The upload didn't work, you might have to wait a bit longer.

Monsurat-Onabajo:
> @Monsurat-Onabajo The upload didn't work, you might have to wait a bit longer.

I have uploaded it again

OriolAbril:
@Monsurat-Onabajo @BerylKanali if you don't mind waiting we can take a look this Thursday, maybe it is easier to solve in the sync session

BerylKanali:
> @Monsurat-Onabajo @BerylKanali if you don't mind waiting we can take a look this Thursday, maybe it is easier to solve in the sync session

@OriolAbril Sure. We can look at it then.

BerylKanali:
@fonnesbeck I made additional changes with the guidance of @OriolAbril

examples/howto/linear_model_v1.nc
OriolAbril, line 1: this file should not be commited, can you delete it so it is not included with the PR?

BerylKanali: Done

OriolAbril (approved):

Monsurat-Onabajo:
> @fonnesbeck I made additional changes with the guidance of @OriolAbril cc: @Monsurat-Onabajo

Goodjob beryl

PR #542, opened 2023-04-07
Try older theme version

OriolAbril:
Lets see if this changes anything


OriolAbril:
Haven't checked all the notebooks but it seems to be working

PR #565, opened 2023-08-03
including ModelBuilder intro notebook

michaelraczycki:
This PR includes the introduction of new module of PyMC-experimental, ModelBuilder. 

----
:books: Documentation preview :books:: https://pymc-examples--565.org.readthedocs.build/en/565/


drbenvincent:
Hi @michaelraczycki. Looks like this could be pretty cool. Before diving in to a proper review, a few changes are needed:

* I'd pick a simpler name for the notebook, ideally no spaces. This will be useful if/when it comes to using sphinx to reference between notebooks in the docs.
* Placing the notebook inside a new folder `model_builder` will create a new section in the docs. I don't think we need a complete section on the module builder. Instead, I'd recommend putting it in the `howto` folder/section.
* Notebooks for `pymc-examples` have a reasonably hefty set of formatting and style requirements. There's a very useful guide here https://www.pymc.io/projects/docs/en/latest/contributing/jupyter_style.html which in turn has links to a video and a template notebook.

Just give me a shout when it's ready for re-review.

michaelraczycki:
@drbenvincent I think that all the formatting should be there now

twiecki:
Why not update https://github.com/pymc-devs/pymc-examples/blob/main/examples/howto/model_builder.ipynb ?

twiecki:
@michaelraczycki There's 2 NBs here, is that on purpose? Also, the pre-commit should generate a myst file that should be included here too.

michaelraczycki:
you're right, for some reason the deletion of the model_builder folder and it's notebook didn't get committed. It's fixed now

PR #567, opened 2023-08-23
Fix Stucchio URL

maresb:
The backslashes are appearing in the actual URL

Here's the actual page:

https://www.pymc.io/projects/examples/en/latest/case_studies/bayesian_ab_testing_introduction.html#id79 

Note that this old PyMC3 page is much higher on Google:
https://www.pymc.io/projects/docs/en/v3/pymc-examples/examples/case_studies/bayesian_ab_testing.html

Presumably the backslashes were put there for a reason. Maybe something changed with respect to the parsing though? (I'm just saying that I hope fixing this doesn't break something else.)

I'm ignoring the template since I'm modifying the `.bib` file and not the notebook file.

# {Insert Description}

### Helpful links

----
:books: Documentation preview :books:: https://pymc-examples--567.org.readthedocs.build/en/567/

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -15,5 +15,5 @@ jobs:
     - uses: actions/setup-python@v2
     - uses: actions/setup-node@v2
       with:
-        node-version: '15'
+        node-version: '18'
     - uses: pre-commit/action@v2.0.0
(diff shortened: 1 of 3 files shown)


OriolAbril:
It looks like the auto sorting and cleanup of the bibtex file is to blame for this. Can you try updating its version and see if it no longer happens? It wants to undo all the changes you did 😅 

Otherwise we'd need to update the pre-commit configuration to use the `--no-encode-urls` flag. Ref: https://github.com/FlamingTempura/bibtex-tidy#cli

maresb:
Thanks @OriolAbril for tracking this down! (I forgot about pre-commit and didn't notice that failure.) Updating the hook (and the node version) seemed to do the trick, and I fixed another similarly broken URL in the process.

OriolAbril (approved):

OriolAbril:
Thanks, that was a great catch!

Issue #554, opened 2023-07-04
Interventions versus counterfactuals


drbenvincent:
# Notebook proposal
**Title**: Interventions versus counterfactuals

## Why should this notebook be added to pymc-examples?
There is a lot of excitement about causal inference at the moment. Interventions and Counterfactuals map on to Pearl's Rung 2 and Rung 3 of the causal ladder. So having a notebook that succinctly explains the concepts and how to implement these in PyMC would be valuable. 

I'd aim to introduce the concepts, then illustrate with a practical example. This could perhaps be time-series based in order to emphasise the retrospective vs prospective aspects of intervention and counterfactuals.

Suggested categories:
* Level: Intermediate
* [Diataxis](https://diataxis.fr/) type: Explanation or How-To

## Related notebooks
This notebook would compliment the range of other causal focussed notebooks that I've put together:
<img width="605" alt="Screenshot 2023-07-04 at 20 39 56" src="https://github.com/pymc-devs/pymc-examples/assets/6765047/2010a975-9e0c-4301-8d16-5520ee4ae89b">
It could be a good sequel to the [Interventional distributions and graph mutation with the do-operator](https://www.pymc.io/projects/examples/en/latest/causal_inference/interventional_distribution.html) example, but it wouldn't really overlap with it.

# References
* Pearl

PR #569, opened 2023-08-25
Counterfactual generation using pymc do-operator example notebook

shekharkhandelwal1983:
# This notebook provides a detailed steps required to generate counterfactuals, and demonstrates the newly introduced do-operator capabilities. And how it can be used to achieve Causality.

### Helpful links


drbenvincent:
Might overlap with #554 which I'm working on with @juanitorduz. Not necessarily the end of the world though - there's a clear opportunity for stuff on this topic.  Would like to review this - but just became a father (yesterday!) so there might be a minor delay

twiecki:
What I really like about this NB is how it lays out the new workflow with creating the skeleton model first. Maybe that should be the focus.

aloctavodia:
In case this is still useful, I updated this to work with new versions of PyMC and ArviZ. If not useful, feel free to close.

PR #597, opened 2023-11-17
move BART case studies into seperate category

daniel-saunders-phil:
The case studies section of the examples should be subdivided into more specific categories. BART is a distinct type of modeling rather than a case study. We have enough notebooks for it now that it can stand on its own.

- [x] update thumbnail_extractor
- [x] standardize titles BART vs bart

----
:books: Documentation preview :books:: https://pymc-examples--597.org.readthedocs.build/en/597/


juanitorduz:
@aloctavodia does this have any effect on BART docs (because of the paths?)

twiecki (approved):

OriolAbril:
we can set up redirects to preserve the old paths, but bart docs should use the `(target)=` at the top of the notebook to reference them for things like that; if the target doesn't change, the cross-reference is still valid

OriolAbril:
add redirects here: https://github.com/pymc-devs/pymc-examples/blob/main/examples/conf.py#L155; keys are old/unexistent paths (redirect origin), values are existing paths (redirect endpoint)

twiecki:
@OriolAbril I don't think we need to add redirects, if people want stable links they should reference a specific version of the docs.

aloctavodia:
Thanks, @daniel-saunders-phil  for the PR. I am working on BART right now I will update the paths if necessary.

PR #603, opened 2023-11-17
GP-Smoothing Notebook to V5

juanitorduz:
Update to V5

----
:books: Documentation preview :books:: https://pymc-examples--603.org.readthedocs.build/en/603/


twiecki (approved):

twiecki:
Weird NB, doesn't even use the GP submodule. Should we even keep it around?

juanitorduz:
> Weird NB, doesn't even use the GP submodule. Should we even keep it around?

Indeed! It is using `GaussianRandomWalk` to smooth the data. It is interesting but I am not sure if important enough to keep it. I do not have strong opinions. (It was more of a low hanging fruit)

twiecki:
Let's delete it.

twiecki:
Still merging so we have the latest in the history.

PR #604, opened 2023-11-17
Move blackbox likelihoods to howto. Move samplers to samplers.

twiecki:
Not sure what this is about:

```
Check no references that should be sphinx cross-references are urls...........................Failed
- hook id: no-references-as-links
- exit code: 1

examples/howto/blackbox_external_likelihood_numpy.ipynb:370:    "But, what if we don't know the analytical form. If our model/likelihood is purely Python and made up of standard maths operators and Numpy functions, then the [autograd](https://github.com/HIPS/autograd) module could potentially be used to find gradients (also, see [here](https://github.com/ActiveState/code/blob/master/recipes/Python/580610_Auto_differentiation/recipe-580610.py) for a nice Python example of automatic differentiation). But, if our model/likelihood truly is a \"black box\" then we can just use the good-old-fashioned [finite difference](https://en.wikipedia.org/wiki/Finite_difference) to find the gradients - this can be slow, especially if there are a large number of variables, or the model takes a long time to evaluate. Below, a function to find gradients has been defined that uses the finite difference (the central difference) - it uses an iterative method with successively smaller interval sizes to check that the gradient converges. But, you could do something far simpler and just use, for example, the SciPy [approx_fprime](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.approx_fprime.html) function.\n"

```


OriolAbril (commented):
Could be merged as is, but I noticed a couple issues that we should try to tackle at some point:

* there are still several links to `deeplearning.com` from the old theano docs
* there are now multiple broken references all around. Here for example I have noticed references to the cython notebook and the custom distribution one, which are gone

twiecki:
Cool, I'll merge and open an new issue.

twiecki:
Issue: https://github.com/pymc-devs/pymc-examples/issues/605

PR #610, opened 2023-11-27
change level category look

### LINKED ISSUE #607: Change category items for beginner, intermediate, advanced from birds to text
Currently there are three birds which users fails to understand the meaning of:
image

Would be better to change to text. However, looking at the code: https://github.com/pymc-devs/pymc-examples/blob/main/_templates/postcard_categories.html#L38 I couldn't figure out how to change it the correct way. @OriolAbril do you know?

OriolAbril:
Closes #607

----
:books: Documentation preview :books:: https://pymc-examples--610.org.readthedocs.build/en/610/

### DIFF
--- a/_templates/postcard_categories.html
+++ b/_templates/postcard_categories.html
@@ -11,9 +11,9 @@
   {% endif %}
 {%- endmacro -%}
 
-{%- macro category_item(name, number, active) -%}
+{%- macro category_item(name, active) -%}
 <a href="{{ pathtocategory(name) }}" class="{{ setcolorclass(active) }}" title="{{ name }}">
-  <b>{{ number }}</b>
+  <b>{{ name }}</b>
 </a>
 {%- endmacro -%}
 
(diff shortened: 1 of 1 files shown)


OriolAbril:
It also adds level at the beginning of the row:

imatge

It would also be possible to use words, but it will take more time to get the css so all the words fit in a single row. If someone can do it please go for it. I think the PR at least will show how to update the jinja template

ricardoV94:
Suggestion: Keep the icons (I think they are cute), but add label afterwards:
bird: Beginner
bird flying: Intermediate
dragon: Advanced

Second suggestion: Move the 4 label links: How-to/Tutorial/Reference... to a less prominent location? I am skeptical they are a useful navigation aid to users.

twiecki:
image

OriolAbril:
can we keep the type of documentation category even if somewhere else?

twiecki:
@OriolAbril They still show up here: https://www.pymc.io/projects/examples/en/latest/blog/category.html or where do you think we should have them?

twiecki:
For me the difference between tutorial, howto, explanation is not clear.

twiecki:
> The birds are gone :( I liked them

What did you like about them?

twiecki:
That they were.

Issue #617, opened 2023-12-07
Slight Mis-Interpretation of Discrete Choice Theory


mitch-at-orika:
Discrete Choice and Random Utility Models:
[**Notebook url**](https://www.pymc.io/projects/examples/en/latest/generalized_linear_models/GLM-discrete-choice_models.html):

## Issue description
Firstly thanks for putting this up, its a great effort @NathanielF something that took me ages to figure out and your example will be a great resource for others going forward. I wanted to highlight that In discrete choice models you dont typically set the whole utility to zero for one alternative you just set one of the contants to zero.
Refer to page 21 of [Train Book Ch2](https://eml.berkeley.edu/books/choice2nd/Ch02_p9-33.pdf) for why - its about the identification issue.

I do acknowledge though that it isnt uncommon to do it like you have done as per this example: [BAPs multinomial_ppcs](https://github.com/aloctavodia/BAP/blob/master/extras/multinomial_ppcs.ipynb) that does set everything to zero for one class. 

My issue would be that the model is then ignoring the Xs in that alternative. Also, as you articulate well, discrete choice is being nuanced in how the utility is expressed per alternative and saying one must be zero is killing that. 

## Expected output / Proposed solution
I think the first example (no contants) should just have utility like other alternatives. 

then second one where you bring in contants you can either just 
u4 = 0 + beta_ic * wide_heating_df["ic.gr"] + beta_oc * wide_heating_df["oc.gr"]
or concatenate a zero onto the alpha if you want to do it as a tensor op like Alex did in the Multinomial PPCs example.

Thanks again.

NathanielF:
Thanks for thinking through this @mitch-at-orika ! 

Would you like to open a PR demonstrating the difference in the specification? In particular it'd be interesting to discuss differences in interpretation where relevant.

mitch-at-orika:
I dont do this PR on public repos often... I hope I got it all right but there is a PR up there now. Results of the option with no constant are better and results of the example with constants are ~ same maybe a bit worse. The marginal rate of substitition calc goes from 2 -> 0.7. Tried comparing with the mlogit example I think they are both a bit different. Hope this helps, let me know if I need to do the PR differently better. Also keen to talk through implications. Cheers, Mitch

NathanielF:
Just FYI @mitch-at-orika finally updating the PR here with the suggested changes: https://github.com/pymc-devs/pymc-examples/pull/781

PR #618, opened 2023-12-08
New discrete choice

mitch-at-orika:
# Testing one alternative utility not zero (just the constant)

Issue #617

### DIFF
--- a/examples/generalized_linear_models/GLM-discrete-choice_models.myst.md
+++ b/examples/generalized_linear_models/GLM-discrete-choice_models.myst.md
@@ -5,9 +5,9 @@ jupytext:
     format_name: myst
     format_version: 0.13
 kernelspec:
-  display_name: pymc_examples_new
+  display_name: Python 3 (ipykernel)
   language: python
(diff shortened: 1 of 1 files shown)


NathanielF:
Thanks for the PR @mitch-at-orika! Sorry it took me so long to get back to you - Christmas has been waaay too busy.  I've added a few comments above. Mostly about deciding how consistent we want to be in updating the model specification and making sure the rest of the text reflects the new model specification.

NathanielF:
On the pre-commit checks it looks like you just need to reset the notebook and run all cells from the top.

mjbaske:
@NathanielF can this possible be resurected of got a few things about model eval and macfaddens rho squared I would like to add but dont know if the branch is now way too out of date?

NathanielF:
Hi @mjbaske !

There has not been any changes to the  notebook that I know of, but the pymc examples repo itself has likely been updated. I think you would be better off starting a new branch if the updates you envision are different from the changes started here. Just to keep things cleaner.

PR #631, opened 2024-02-02
Update blackbox likelihood example

ricardoV94:
## Changes
- Make Op more generic (accepts distinct inputs, returns vectorized loglikelihood), for better transfer learning
- Build Ops more iteratively, and show useful debugging strategies
- Use CustomDist instead of Potential (added a short section at the end showing Potential)
- Actually make use of finite differences as mentioned in the text
- Remove mention of old Cython example and reference new JAX example
- Fix some cross-references (unclear success)

## Checklist

----
📚 Documentation preview 📚: https://pymc-examples--631.org.readthedocs.build/en/631/

Link to NB: https://pymcio--631.org.readthedocs.build/projects/examples/en/631/howto/blackbox_external_likelihood_numpy.html


ricardoV94:
Is the pre-commit.ci failure to build known?

examples/howto/blackbox_external_likelihood_numpy.myst.md
```diff
@@ -175,70 +167,146 @@ x = np.linspace(0.0, 9.0, N)
...
+
+```{code-cell} ipython3
+my_loglike(mtrue, ctrue, sigma, x, data)
+```
+
+### Model definition
+
++++
+
+Now, let's use this Op to repeat the example shown above. To do this let's create some data containing a straight line with additive Gaussian noise (with a mean of zero and a standard deviation of `sigma`). For simplicity we set {class}`~pymc.Uniform` prior distributions on the gradient and y-intercept. As we've not set the `grad()` method of the Op PyMC will not be able to use the gradient-based samplers, so will fall back to using the {class}`pymc.Slice` sampler.
```
OriolAbril, line 210: The pymc.Slice isn't rendered as a link, but the reference is right here and should be left as is. The issue is on the PyMC api docs.

examples/howto/blackbox_external_likelihood_numpy.myst.md
```diff
@@ -175,70 +167,146 @@ x = np.linspace(0.0, 9.0, N)
...
 ```
 
-+++ {"jp-MarkdownHeadingCollapsed": true}
-
 ## PyTensor Op with grad
 
-What if we wanted to use NUTS or HMC? If we knew the analytical derivatives of the model/likelihood function then we could add a {func}`grad() method <pytensor:creating_an_op>` to the Op using that analytical form.
+What if we wanted to use NUTS or HMC? If we knew the analytical derivatives of the model/likelihood function then we could add a {func}`grad` to the Op using existing PyTensor operations.
+
+But, what if we don't know the analytical form. If our model/likelihood, is implemented in a framework that provides automatic differentiation (just like PyTensor does), it's possible to reuse their functionality. This {ref}`related example <wrapping_jax_function> shows how to do this when working with JAX functions.
```
OriolAbril, line 270:
```suggestion
But, what if we don't know the analytical form. If our model/likelihood, is implemented in a framework that provides automatic differentiation (just like PyTensor does), it's possible to reuse their functionality. This {ref}`related example <wrapping_jax_function>` shows how to do this when working with JAX functions.
```

backtick missing breaks rendering

examples/howto/blackbox_external_likelihood_numpy.myst.md
```diff
@@ -409,23 +536,31 @@ az.plot_pair(idata_grad, **pair_kwargs, ax=ax)
...
-az.plot_pair(idata, **pair_kwargs, ax=ax);
+az.plot_pair(idata_pure, **pair_kwargs, ax=ax);
 ```
 
-We can now check that the gradient Op works as expected. First, just create and call the `LogLikeGrad` class, which should return the gradient directly (note that we have to create a {ref}`PyTensor function <pytensor:creating_an_op>` to convert the output of the Op to an array). Secondly, we call the gradient from `LogLikeWithGrad` by using the {func}`grad` function. Finally, we will check the gradient returned by the PyMC model for a Normal distribution, which should be the same as the log-likelihood function we defined. In all cases we evaluate the gradients at the true values of the model function (the straight line) that was created.
+## Using a Potential instead of CustomDist
+
++++
+
+In this section we show how a {class}`pymc.Potential` can be used to implement a black-box likelihood a bit more directly than when using a a {class}`pymc.CustomDist`.
```
OriolAbril, line 546:
```suggestion
In this section we show how a {func}`pymc.Potential` can be used to implement a black-box likelihood a bit more directly than when using a a {class}`pymc.CustomDist`. 
```

Potential is actually a function, so {class} wouldn't work for the reference. Similarly to pymc.Slice though, it won't work yet either until we revert and fix pymc api docs, which currently document it at `pymc.model.core.Potential`. Same goes for all other potential references, they should be `` {func}`pymc.Potential` ``

OriolAbril:
> Is the pre-commit.ci failure to build known?

I have no idea, I haven't tracked the latest PRs and wasn't even aware we had it (or what it does).

> I am really really hesitant to use finite difference gradients in an example, because it suggests that this is a good idea. I think it would be better to show the pencil pushing to get analytic gradients.

Similar thinking on my end. If it really is blackbox, I think the recommendation should go more towards use samplers that don't require gradients (several of which are available in PyMC already so it doesn't complicate anything really). Here however it isn't so we can go ahead and also illustrate how to extend the Op with the grad method.

jessegrabowski (commented):
I added some thoughts on NB review

ricardoV94:
<p>I think we need a proper notebook on CustomDist with dist behavior, probably a core notebook, not even pymc-examples.</p><p> </p><p>This tutorial has always been about a black-box likelihood, original a complicated cython function, that got removed in the last hackathon because it was too complicated for devs to maintain. Now we have this baby numpy alternative here, which I think it's fine.</p><p> </p><p>The different between a blackbox Likelihood/model and individual Ops is thin. I don't want to delve into that here because it requires more PyTensor exposition. I just wanted to make this NB more useful than it was before not to change the pedagogy.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924868825'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>I think contrived example is fine. Before the example failed because the parameters were passed as a vector and there was an indexing operation. Only a sufficiently complex example will be non-contrived, and that will be necessarily less readable. In the JAX example I did try this by writting a HMM with jax.scan and making use of vmap (of course you can do it in PyTensor, that was before <code>vectorize</code> was added).</p><p> </p><p>There was already pushback there that the example was too complex and detracted from the point. </p><p> </p><p><a href="https://www.pymc.io/projects/examples/en/latest/howto/wrapping_jax_function.html" target="_blank">https://www.pymc.io/projects/examples/en/latest/howto/wrapping_jax_function.html</a></p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924868869'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>Yeah, definitely</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924868923'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>Agree</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924868977'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>I know where you are coming from I also see users trying to get out of PyTensor/math too soon. </p><p> </p><p>Still for simplicity sake I would keep it. I think the example with the closed form solution was less informative. Also finite differences are a great gradient debugging tool (and what verify_grad does) so it's nice to show it. I can emphasize that analytic expressions are usually preferable more strongly</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924869140'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>May be good to emphasize that we could black box the model part only, and still use a <code>pm.Normal</code> likelihood, or we can black-box both the model and likelihood (and even priors). The way the original example was written already leads itself to it, by separating <code>my_model</code> and <code>my_loglike</code>  </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924868825'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>Thanks</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924869177'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>That's a big part of what I wanted to show in the updated NB. Users always ask about multiple inputs / outputs when referring to this NB / attempting their own Op</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924869217'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>Is the part mentioned earlier not enough?</p><p> </p><pre class="ql-syntax" spellcheck="false">Another issue is that if you want to be able to use the gradient-based step samplers like {term}<code>NUTS</code> and {term}<code>Hamiltonian Monte Carlo</code>, then your model/likelihood needs a gradient to be defined. If you have a model that is defined as a set of PyTensor operators then this is no problem - internally it will be able to do automatic differentiation - but if your model is essentially a "black box" then you won't necessarily know what the gradients are.
</pre>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924868923'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>On revisiting this, note sure we can do much with the <code>make_node</code> type-hints. I wrote it to accept Numpy inputs as well as PyTensor objects (with the call to <code>as_tensor</code>), which I find makes life much easier for debugging. So I have only added the Apply as output type hint. <code>perform</code> is also odd because it doesn't return anything, but writes in place in the lists. Will type-hints to those anyway    </p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924868977'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>Fixed</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924869036'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>Fixed</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924869087'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>Rephrased text to be less positive and added an explicit caution section</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924869140'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ricardoV94:
<p>Fixed</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/631/discussion/#comment-1924869260'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

examples/howto/blackbox_external_likelihood_numpy.myst.md
ricardoV94: Fixed

examples/howto/blackbox_external_likelihood_numpy.myst.md
ricardoV94: Fixed, sneaky PyMC capitalized functions

OriolAbril (approved):

PR #652, opened 2024-04-01
Add a mamba-compatible env file

AlexAndorra:
There was no YML env file -- only `.txt` ones -- so I added one to make install with mamba seamless.
Ready to merge

----
📚 Documentation preview 📚: https://pymc-examples--652.org.readthedocs.build/en/652/

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,29 @@
+name: pymc-examples
+channels:
+- conda-forge
+- defaults
+dependencies:
+# Base dependencies
+- black
+- codespell
+- ipython
+- jupyterlab
+- notebook
+- pandas
+- pre-commit>=2.8.0
+- pymc
+- watermark
+# Extra dependencies for docs build
+- ablog>=0.11
+- matplotlib
+- myst-nb
+- sphinx-codeautolink
+- sphinx>=5
+- pymc-sphinx-theme==0.14
+- sphinx-copybutton
+- sphinx-design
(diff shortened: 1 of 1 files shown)


OriolAbril:
Can env files be combined when installing/creating envs? If so I would create two, one with the dependencies needed to write notebooks, another with the dependencies needed to build the website. Other than ipython/jupyter the rest are all different for each task and many people will only work on one of the two (generally writing notebooks) without doing the other at all.

If they can't be combined I'd use comments to split dependencies into three sections: base, writing and building.

AlexAndorra:
Thanks @OriolAbril , your comment led me to get my mind blown: you can actually [create envs](https://docs.conda.io/projects/conda/en/latest/commands/create.html) with mamba... with text files -- it doesn't require YML files 🤯 

Something like this will do the trick:
`mamba create -n pyex --file=requirements-docs.txt --file=requirements-write.txt --channel="conda-forge"`

So I guess I'll just close this PR 😅

AlexAndorra:
Thanks @OriolAbril , your comment led me to get my mind blown: you can actually [create envs](https://docs.conda.io/projects/conda/en/latest/commands/create.html) with mamba... with text files -- it doesn't require YML files 🤯 

Something like this will do the trick:
`mamba create -n pyex --file=requirements-docs.txt --file=requirements-write.txt --channel="conda-forge"`

So I guess I'll just close this PR 😅

PR #660, opened 2024-05-07
Bound MyST NB to fix the compressed images

juanitorduz:
The images of the gallery are being compressed by a change added in the latest release of MyST-NB. See https://github.com/pymc-labs/CausalPy/issues/327#issuecomment-2096256657 

This is a temporary fix until they resolve it (see https://github.com/pymc-labs/pymc-marketing/pull/667)

----
📚 Documentation preview 📚: https://pymc-examples--660.org.readthedocs.build/en/660/

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -1,6 +1,6 @@
 ablog>=0.11
 matplotlib
-myst-nb
+myst-nb<=1.0.0
 sphinx-codeautolink
 sphinx>=5
 pymc-sphinx-theme==0.14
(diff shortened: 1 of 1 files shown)


requirements-docs.txt
```diff
@@ -1,6 +1,6 @@
 ablog>=0.11
 matplotlib
-myst-nb
+myst-nb<=1.0.0
```
OriolAbril, line 3:
```suggestion
myst-nb!=1.1
```

There is already a PR open with the fix, so I think we can do this so we don't have to undo the pin, simply indicate this specific version is faulty. What do you think? Either way should work though, just trying to ponder which is more probable to end up with less future work

OriolAbril (approved):

juanitorduz:
Thanks @OriolAbril ! When is green we can merge 🟢🤜🤛

PR #665, opened 2024-06-01
Update HSGP with changes from PyMC #7342

bwengals:
- Update usage of `pm.gp.hsgp_approx.approx_hsgp_hyperparams`
- Replace `Xs` input with `X` on `.prior_linearized` 
- Add reference to Stan prior rec's for student-t degrees of freedom
- Tweak priors on example 2 for better results

----
📚 Documentation preview 📚: https://pymc-examples--665.org.readthedocs.build/en/665/

### DIFF
--- a/examples/gaussian_processes/HSGP-Basic.myst.md
+++ b/examples/gaussian_processes/HSGP-Basic.myst.md
@@ -5,9 +5,9 @@ jupytext:
     format_name: myst
     format_version: 0.13
 kernelspec:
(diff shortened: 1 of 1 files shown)


AlexAndorra (changes requested):
Thanks @bwengals ! I think there is a typo in the computation of X_center for the Gram matrix. Otherwise it looks all good!

AlexAndorra (approved):

Issue #659, opened 2024-05-03
Update data container notebook to reflect changes to `pm.Data`


jessegrabowski:
https://github.com/pymc-devs/pymc/pull/7047 depreciates `pm.ConstantData` and `pm.MutableData` (again), so the [data container notebook](https://www.pymc.io/projects/examples/en/latest/fundamentals/data_container.html) no longer reflects best practices. It should be updated to:

1. Remove the explainer about the differences between `MutableData` and `ConstantData`, since it no longer exists, and;
2. Use `pm.Data` everywhere in the applied examples

Dekermanjian:
Hey, would it be alright if I worked on this?

PR #676, opened 2024-07-28
updated data_container.ipynb to reflect changes to pm.Data containers in pymc v5.16

Dekermanjian:
related to [#659](https://github.com/pymc-devs/pymc-examples/issues/659)

The data container notebook is out of date. It uses the deprecated `pm.ConstantData` and `pm.MutableData` methods.
The changes I made were:

- Update the introduction to reflect the change to using `pm.Data` but keeping a section that refers to the deprecated `pm.MutableData` and `pm.ConstantData` in case a new user sees that in example code elsewhere. 
- Replace all instances of `pm.ConstantData` and `pm.MutableData` with `pm.Data` in the code and in the markdown text
- Replace the instance of mutable `coords` with `coords` to reflect that all coordinates are now mutable and explain that fact in the preceding markdown text
- I removed some text that no longer applied like checking the type of data container twice previously for each type of data container
- I made a couple of typo corrections


examples/fundamentals/data_container.myst.md
```diff
@@ -58,19 +58,17 @@ This notebook will illustrate each of these benefits in turn, and show you the b
 
 +++
 
-## Types of Data Containers
+## Current and past states of Data Containers
```
twiecki, line 61: I think this whole section can be deleted and maybe replaced with an info box of your last comment.

examples/fundamentals/data_container.myst.md
```diff
@@ -99,20 +97,20 @@ Furthermore, inside `idata`, PyMC has automatically saved the observed (endogeno
 idata.observed_data
 ```
 
-In this next model, we create a `pm.ConstantData` containers to hold the observations, and pass this container to the `observed`. We also make a `pm.ConstantData` container to hold the `x` data:
+In this next model, we create a `pm.Data` container to hold the observations, and pass this container to the `observed`. We also make a `pm.Data` container to hold the `x` data:
```
twiecki, line 100:
```suggestion
In this next model, we create a {class}`pm.Data` container to hold the observations, and pass this container to the `observed`. We also make a {class}`pm.Data` container to hold the `x` data:
```

examples/fundamentals/data_container.myst.md
```diff
@@ -99,20 +97,20 @@ Furthermore, inside `idata`, PyMC has automatically saved the observed (endogeno
...
+    y_data = pm.Data("y_data", y)
     beta = pm.Normal("beta")
     mu = pm.Deterministic("mu", beta * x_data)
     sigma = pm.Exponential("sigma", 1)
     obs = pm.Normal("obs", mu=mu, sigma=sigma, observed=y_data)
     idata = pm.sample(random_seed=RANDOM_SEED)
 ```
 
-Because we used a `pm.ConstantData` container, the data now appears on our probabilistic graph. It is downstream from `obs` (since the `obs` variable "causes" the data), shaded in gray (because it is observed), and has a special rounded square shape to emphasize that it is data. We also see that `x_data` has been added to the graph.
+Because we used a `pm.Data` container, the data now appears on our probabilistic graph. It is downstream from `obs` (since the `obs` variable "causes" the data), shaded in gray (because it is observed), and has a special rounded square shape to emphasize that it is data. We also see that `x_data` has been added to the graph.
```
twiecki, line 113:
```suggestion
Because we used a `pm.Data` container, the data now appears in our probabilistic graph. It is downstream from `obs` (since the `obs` variable "causes" the data), shaded in gray (because it is observed), and has a special rounded square shape to emphasize that it is data. We also see that `x_data` has been added to the graph.
```

examples/fundamentals/data_container.myst.md
twiecki: Change everywhere we use `pm.Data`.

examples/fundamentals/data_container.myst.md
Dekermanjian: I agree, I will make the change.

examples/fundamentals/data_container.myst.md
Dekermanjian: Thank you for clarifying that for me. I was not sure if the `{class}` was only supposed to appear the first time we introduce the class. I will make the changes to fix it.

examples/fundamentals/data_container.myst.md
Dekermanjian: Yup, I missed that. Thank you, for catching it. I will fix it.

examples/fundamentals/data_container.myst.md
OriolAbril: It should be one of ``{class}`pymc.Data` `` (will render pymc.Data as link) or ``{class}`~pymc.Data` `` (will render only Data as link). Sphinx doesn't know about `pm`. I prefer the 1st one.

Dekermanjian: @twiecki @OriolAbril Would it be preferable to change all `pm.Data` to `{class}pymc.Data` or is it preferred that I go back to keeping it as `pm.Data` without `{class}`? At the start of the notebook, there is a reference to `{class}pymc.Data`.

OriolAbril: I personally prefer `pymc.Data` as a link, but I don't really care much either way

Dekermanjian: Okay, I went ahead and changed it so that it is {class} `pymc.Data` everywhere in the markdown text.

examples/fundamentals/data_container.myst.md
```diff
@@ -58,19 +58,15 @@ This notebook will illustrate each of these benefits in turn, and show you the b
 
 +++
 
-## Types of Data Containers
-
- PyMC offers two data containers, depending on your needs: {func}`pymc.ConstantData` and {func}`pymc.MutableData`. Both will help you visualize how data fits into your model, store the data in an `InfereceData` for reproducibility, and give access to labeled dimenions. As the names suggest, however, only `MutableData` allows you to change your data. When `X` is `MutableData`, this enables out-of-sample inference tasks. When `y` is `MutableData`, it allows you to reuse the same model on multiple datasets to perform parameter recovery studies or sensitivity analysis.  These abilities do, however, come with a small performance cost.
- 
- In past versions of PyMC, the only data container was `pm.Data`. This container is still available for backwards compatability, but the current best practice is to use either `pm.MutableData` or `pm.ConstantData`. 
+:::{important}
+In past versions of PyMC, there were two types of data containers {func}`pymc.MutableData` and {func}`pymc.ConstantData` these are now deprecated.
```
twiecki, line 62:
```suggestion
In past versions of PyMC, there were two types of data containers {func}`pymc.MutableData` and {func}`pymc.ConstantData`. These have been deprecated as all data containers are mutable now.
```

Dekermanjian: Thank you, for the review @twiecki. I have made the change that you have suggested.

Issue #115, opened 2021-04-05
updating priors


OriolAbril:
https://github.com/pymc-devs/pymc-examples/blob/main/examples/pymc3_howto/updating_priors.ipynb
Reviewers:

> The sections below may still be pending. If so, the issue is still available, it simply doesn't
> have specific guidance yet. Please refer to [this overview of updates](https://github.com/pymc-devs/pymc-examples/wiki/Notebook-updates-overview)

## Known changes needed
Changes listed in this section should all be done at some point in order to get this
notebook to a "Best Practices" state. However, these are probably not enough!
Make sure to thoroughly review the notebook and search for other updates.

### General updates
*

### ArviZ related
*

## Changes for discussion
Changes listed in this section are up for discussion, these are ideas on how to improve
the notebook but may not have a clear implementation, or fix some know issue only partially.

### General updates
*

### ArviZ related
*

symeneses:
I would like to work on this issue 🙋🏽‍♀️

OriolAbril:
@symeneses moving the discussion from https://github.com/pymc-devs/pymc-examples/issues/394#issuecomment-1192300850 to here.

The approach used in the updating priors notebook has multiple problems, the most critical one being that it uses univariate priors, which is generally not the case and can't be extended or used in other cases. I wrote a bit on this in one of my blogposts: https://oriolabrilpla.cat/python/arviz/pymc/xarray/xarray-einstats/2022/05/25/too-eager-reduction.html#Univariate-priors. 

I'd recommend rewriting it not from scratch but nearly so that it uses [prior_from_idata](https://www.pymc.io/projects/experimental/en/latest/api_reference.html#pymc_experimental.utils.prior.prior_from_idata) from the pymc-experimental library. It won't depend on pm.Interpolated anymore and should not be blocked until the issues with this are fixed. Let me know how it sounds. If you prefer it we could also leave this for another future PR.

symeneses:
I think we can update this notebook using `pm.Interpolated` if this is a `good example` of how to use `pm.Interpolated`, and create after a similar one using the approach you suggested. Otherwise, we can update the notebook as you suggested.

OriolAbril:
[...]

PR #692, opened 2024-08-04
update notebook: Update Priors

juanitorduz:
Update https://www.pymc.io/projects/examples/en/latest/howto/updating_priors.html to PyMC 5

Related to https://github.com/pymc-devs/pymc-examples/issues/115

It was still using PyMC3 and Theano :/

----
📚 Documentation preview 📚: https://pymc-examples--692.org.readthedocs.build/en/692/


examples/howto/updating_priors.myst.md
```diff
@@ -10,56 +10,61 @@ kernelspec:
   name: python3
 ---
 
-# Updating priors
+(updating-priors)=
+# Updating Priors
+
+:::{post} January, 2017
+:tags: priors
+:category: intermediate, how-to
```
twiecki, line 18: Can we find who the original author was?

examples/howto/updating_priors.myst.md
```diff
@@ -10,56 +10,61 @@ kernelspec:
...
+:::
 
 +++
 
-In this notebook, I will show how it is possible to update the priors as new data becomes available. The example is a slightly modified version of the linear regression in the [Getting started with PyMC3](https://github.com/pymc-devs/pymc3/blob/master/docs/source/notebooks/getting_started.ipynb) notebook.
+In this notebook, we will show how it is possible to update the priors as new data becomes available.
```
twiecki, line 23: I would add a note that this is actually a bit sketchy as the MC error can creep in. I think the last plot actually shows that as some posteriors jump.

examples/howto/updating_priors.myst.md
juanitorduz: No, I could not track the original author via git blame.

examples/howto/updating_priors.myst.md
juanitorduz: +1

OriolAbril:
I'll try to track the original author. I think this is a very nice usage example for Interpolated. But we should be a bit more clear this isn't really a good idea to do in practice not only for the posterior->kde part, but mostly because Interpolated distributions used as priors are unidimensional and uncorrelated, so even if a perfect fit marginally they don't really incorporate all the information we have from the previous posterior into the model, especially when posterior variables are correlated. I wrote a bit on this a while ago https://oriolabrilpla.cat/en/blog/posts/2022/too-eager-reduction.html#univariate-priors

It might also be nice to show or maybe only mention the functionality in https://www.pymc.io/projects/experimental/en/latest/generated/pymc_experimental.utils.prior.prior_from_idata.html which should probably work better for correlated posteriors close to normal, but probably worse than this approach for not correlated but far from normal (with no clear transformation) posteriors.

juanitorduz:
Thanks for the feedback @twiecki and @OriolAbril . In https://github.com/pymc-devs/pymc-examples/pull/692/commits/cd0b93b917ba4358c4aaaf91c0ad72ffb650de86, I added some words of caution (as suggested) and some additional references. I think we should update this notebook as this topic often comes with new practitioners. If we feel the approach can do more harm than good, we can also think about deleting the notebook. (I think we should keep it with these words of caution, but I am easy).

Also, I haven't been able to track the original author since the commit history started after the pymc3 migration.

juanitorduz:
> Original PR: [pymc-devs/pymc#1878](https://github.com/pymc-devs/pymc/pull/1878)

🙏 Wow! 2017!

examples/howto/updating_priors.myst.md
```diff
@@ -116,53 +140,85 @@ traces = [trace]
...
 You can re-execute the last two cells to generate more updates.
 
 What is interesting to note is that the posterior distributions for our parameters tend to get centered on their true value (vertical lines), and the distribution gets thiner and thiner. This means that we get more confident each time, and the (false) belief we had at the beginning gets flushed away by the new data we incorporate.
 
++++
+
+``````{admonition} Not silver bullet
+:class: warning
+Observe that, despite the fact that the iterations seems improving,  toms of them look actually a bit sketchy as the MC error can creep in.
+``````
```
OriolAbril, line 198:
```suggestion
``````{admonition} Not a silver bullet
:class: warning
Observe that, despite the fact that the iterations seems improving, some of them don't look so good, even sometimes it seems it regresses. In addition to reasons noted at the beginning of the notebook, there are a couple key steps in the process where randomness is involved. Thus, things should be expected to improve on average.

1. New observations are random. If in the initial iterations we get values closer to the bulk of the distribuion and then we get several values in a row from the positive tail, the iterations where we have accumulated a couple draws from the tail will probably be biased and "look worse" than previous ones.
2. MCMC is random. Even when it converges, MCMC is a random process, so different calls to `pymc.sample` will return values centered around the exact posterior but not always the same; how large a variation we should expect can be checked with {func}`arviz.mcse`. KDEs also incorporate this often negligible yet present source of uncertainty in the posterior estimates, and so will the generated Interpolated distributions.
``````
```

I tried to rewrite to make it a bit more extensive and clear. Not sure how much of that should be considered "known" for an intermediate notebook but being a very popular notebook I expect it will help

OriolAbril:
I have added some comments. Some are suggestions on the myst file, but the editing process should continue to be work on the ipynb, save then run pre-commit

examples/howto/updating_priors.myst.md
```diff
@@ -10,56 +10,66 @@ kernelspec:
...
-import pymc3 as pm
-import theano.tensor as tt
+import pymc as pm
+import pytensor.tensor as pt
 
-from pymc3 import Model, Normal, Slice, sample
-from pymc3.distributions import Interpolated
 from scipy import stats
-from theano import as_op
+from tqdm.notebook import trange
```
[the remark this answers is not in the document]
juanitorduz: It uses rich now but tqdm is easier to work with for me

juanitorduz:
> I have added some comments. Some are suggestions on the myst file, but the editing process should continue to be work on the ipynb, save then run pre-commit

Amazing! Thanks! 

Indeed! JupyText to the rescue!

OriolAbril (approved):

juanitorduz:
Thank you @OriolAbril !

PR #697, opened 2024-08-31
Updates to Simpon's Paradox notebook

drbenvincent:
This PR updates the Simpon's Paradox notebook in a number of ways:
* `MutableData` -> `Data`
* Updated watermark based on the current style guide
* Added mathematical descriptions of the models
* Improved the hierarchical model. Switched from non-centred to regular parameterisation which gives a simpler graphviz which is more closely aligned to the model maths
* Increased the causal language and have included causal DAGs
* Added `rng` to sampling functions for increased reproducibility
* Significantly simplified potting code
* Added Wilkinson notation (and explanation) for each model. _Hopefully this is correct, but I am not 100% sure._
* Adds a missing notebook reference to the LKJ notebook because I reference to it from this one.
* Added some more data visualisation 
* All models now have a fully pooled observation noise parameter for simplicity.

---

+ [X] Check the notebook is not excluded from any pre-commit check: https://github.com/pymc-devs/pymc-examples/blob/main/.pre-commit-config.yaml

Note: This PR is not related to an issue, I just decided to make these updates in the absence of an issue.

----
📚 Documentation preview 📚: https://pymc-examples--697.org.readthedocs.build/en/697/

### DIFF
--- a/examples/generalized_linear_models/GLM-simpsons-paradox.myst.md
+++ b/examples/generalized_linear_models/GLM-simpsons-paradox.myst.md
@@ -21,7 +21,7 @@ kernelspec:
 
 +++
 
-[Simpson's Paradox](https://en.wikipedia.org/wiki/Simpson%27s_paradox) describes a situation where there might be a negative relationship between two variables within a group, but when data from multiple groups are combined, that relationship may disappear or even reverse sign. The gif below (from the [Simpson's Paradox](https://en.wikipedia.org/wiki/Simpson%27s_paradox) Wikipedia page) demonstrates this very nicely.
+[Simpson's Paradox](https://en.wikipedia.org/wiki/Simpson%27s_paradox) describes a situation where there might be a negative relationship between two variables within a group, but when data from multiple groups are combined, that relationship may disappear or even reverse sign. The gif below (from the Simpson's Paradox [Wikipedia](https://en.wikipedia.org/wiki/Simpson%27s_paradox) page) demonstrates this very nicely.
 
 ![](https://upload.wikimedia.org/wikipedia/commons/f/fb/Simpsons_paradox_-_animation.gif)
 
@@ -41,6 +41,7 @@ import xarray as xr
 ```{code-cell} ipython3
 %config InlineBackend.figure_format = 'retina'
 az.style.use("arviz-darkgrid")
+plt.rcParams["figure.figsize"] = [12, 6]
 rng = np.random.default_rng(1234)
 ```
 
@@ -49,8 +50,6 @@ rng = np.random.default_rng(1234)
 This data generation was influenced by this [stackexchange](https://stats.stackexchange.com/questions/479201/understanding-simpsons-paradox-with-random-effects) question.
 
 ```{code-cell} ipython3
-:tags: [hide-input]
-
 def generate():
     group_list = ["one", "two", "three", "four", "five"]
     trials_per_group = 20
@@ -68,9 +67,8 @@ def generate():
     y = rng.normal(intercept + (x - mx) * slope, 1)
     data = pd.DataFrame({"group": group, "group_idx": subject, "x": x, "y": y})
     return data, group_list
-```
 
-```{code-cell} ipython3
+
 data, group_list = generate()
 ```
 
@@ -137,7 +135,7 @@ with pm.Model() as model1:
 pm.model_to_graphviz(model1)
 ```
 
-### Do inference
+### Conduct inference
 
 ```{code-cell} ipython3
 with model1:
@@ -150,11 +148,11 @@ az.plot_trace(idata1, var_names=["~μ"]);
 
 ### Visualisation
 
-First we'll define a handy predict function which will do out of sample predictions for us. This will be handy when it comes to visualising the model fit.
+First we'll define a handy predict function which will do out of sample predictions for us. This will be handy when it comes to visualising the model fits.
 
 ```{code-cell} ipython3
 def predict(model: pm.Model, idata: az.InferenceData, predict_at: dict) -> az.InferenceData:
-    """Do posterior predictive inference"""
+    """Do posterior predictive inference at a set of out of sample points specified by `predict_at`."""
     with model:
         pm.set_data(predict_at)
         idata.extend(pm.sample_posterior_predictive(idata, var_names=["y", "μ"], random_seed=rng))
@@ -162,6 +160,8 @@ def predict(model: pm.Model, idata: az.InferenceData, predict_at: dict) -> az.In
 ```
 
 ```{code-cell} ipython3
+:tags: [hide-output]
+
 idata1 = predict(
     model=model1,
     idata=idata1,
@@ -234,9 +234,14 @@ The plot on the right shows out posterior beliefs in **parameter space**.
 
 +++
 
-One of the clear things about this analysis is that we have credible evidence that $x$ and $y$ are _positively_ correlated. We can see this from the posterior over the slope (see right hand panel in the figure above).
+One of the clear things about this analysis is that we have credible evidence that $x$ and $y$ are _positively_ correlated. We can see this from the posterior over the slope (see right hand panel in the figure above) which we isolate in the plot below.
 
-+++
+```{code-cell} ipython3
+:tags: [hide-input]
+
+ax = az.plot_posterior(idata1.posterior["β1"], ref_val=0)
+ax.set(title="Model 1 strongly suggests a positive slope", xlabel=r"$\beta_1$");
+```
 
 ## Model 2: Unpooled regression
 
@@ -260,12 +265,16 @@ $$
 \begin{aligned}
 \vec{\beta_0}, \vec{\beta_1} &\sim \text{Normal}(0, 5) \\
 \sigma &\sim \text{Gamma}(2, 2) \\
-\mu_i &= \beta_0[g_i] + \beta_1[g_i] x_i \\
+\mu_i &= \vec{\beta_0}[g_i] + \vec{\beta_1}[g_i] x_i \\
 y_i &\sim \text{Normal}(\mu_i, g_i)
 \end{aligned}
 $$
 
-Where $g_i$ is the group index for observation $i$. So the parameters $\beta_0$ and $\beta_1$ are now length $g$ vectors, not scalars. And the $[g_i]$ acts as an index to look up the group for the $i^{\th}$ observation.
+Where $g_i$ is the group index for observation $i$. So the parameters $\beta_0$ and $\beta_1$ are now length $g$ vectors, not scalars. And the $[g_i]$ acts as an index to look up the group for the $i^\text{th}$ observation.
+
++++
+
+### Build model
 
 ```{code-cell} ipython3
 coords = {"group": group_list}
@@ -284,11 +293,13 @@ with pm.Model(coords=coords) as model2:
     pm.Normal("y", mu=μ, sigma=sigma, observed=data.y, dims="obs_id")
 ```
 
-By plotting the DAG for this model it is clear to see that we now have individual intercept, slope, and variance parameters for each of the groups.
+By plotting the DAG for this model it is clear to see that we now have individual intercept and slope parameters for each of the groups.
 
 ```{code-cell} ipython3
 pm.model_to_graphviz(model2)
 ```
+
+### Conduct inference
 
 ```{code-cell} ipython3
 with model2:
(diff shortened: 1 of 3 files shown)


tomicapretto:
> View / edit / reply to [this conversation](https://app.reviewnb.com/pymc-devs/pymc-examples/pull/697/discussion/#comment-2323359808) on ReviewNB
> 
> _tomicapretto commented on 2024-09-01T14:17:09Z_ ----------------------------------------------------------------
> 
> The formula is correct and the model it builds is equivalent to the one you wrote with distributions above. However, there's not a one-to-one mapping between the distributions above and the formula. The models are equivalent but not exactly the same. With the formula, we're writing the following model
> 
> \begin{aligned}
> 
> p_{0\sigma}, p_{1\sigma} &\sim \text{Gamma}(2, 2) \\ 
> 
> \vec{u_0} &\sim \text{Normal}(0, p_{0\sigma}^2) \\
> 
> \vec{u_1} &\sim \text{Normal}(0, p_{1\sigma}^2) \\
> 
> \beta_0 &\sim \text{Normal} \\ 
> 
> \beta_1 &\sim \text{Normal} \\ 
> 
> \sigma &\sim \text{Gamma}(2, 2) \\ 
> 
> \mu_i &= \beta_0 + \vec{u_0}[g_i] + \beta_1 \cdot x_i + 1\vec{u_1}[g_i] \cdot x_i \\ 
> 
> y_i &\sim \text{Normal}(\mu_i, \sigma) 
> 
> \end{aligned}
> 
> and then `1` maps β 0 , `x` maps β 1 , `(1 | g)` maps u 0 and `(x | g)` maps `u_1$
> 
> This comment in a Bambi discussion can also help
> 
> [bambinos/bambi#808 (comment)](https://github.com/bambinos/bambi/discussions/808#discussioncomment-9242340)

adding the rendered math

image

tomicapretto (commented):
All Wilkinson's formulas are correct, just left a couple of comments though

drbenvincent:
<p>Good idea. Done in an upcoming commit.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/697/discussion/#comment-2323307556'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>So you can actually provide dims in the model without specifying that they are in the coords. For example in Model 2 I provide a coords dict but only provide the group labels. Yet I still have a dimension in the model called obs_id. Because this is not defined, it's easier when it comes to out of sample prediction when shapes change.</p><p> </p><p>I'm happy to explicitly provide all coords in the kwarg, but that will probably require updates in terms of new coords in the out of sample prediction. That might actually be better practice, but let me know if you'd prefer that before I put in the work :)</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/697/discussion/#comment-2323307564'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>In model 3 I was getting divergences and I thought it might help things if I went from unpooled sigmas to pooled. Thought that would actually be ok because the data generating process has identical sigmas for each group.</p><p> </p><p>I just thought I'd keep a single fully pooled parameter in all models to demonstrate you can mix and match.</p><p> </p><p>But could potentially be simpler if Model 2 has unpooled sigmas and Model 3 has hierarchical. Do you think it would be a big improvement to do that?</p><p> </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/697/discussion/#comment-2323307568'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Thanks, this was an error that I didn't catch. Fixed in an upcoming commit</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/697/discussion/#comment-2323359800'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>In terms of divergences: Original notebook used the non-centred parameterisation. The problem that arose was that that makes the DAG and model maths deviate from the regular parameterisation and made it more difficult to explain the connections between the Wilkinson notation, model maths, and pymc model code.</p><p> </p><p>In terms of sigma:  See my reply another of your comments above. Do you think a hierarchy on sigma would reduce divergences? </p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/697/discussion/#comment-2323307575'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Ah, so I think I understand 100%, let me check. The formula block I wrote is equivalent to your formula block. Your's encodes a population mean and group <em>deflections</em> but mine just encodes the absolute group values (not as deflections from population mean).</p><p> </p><p>However, when it comes to comparing to the Wilkinson notation, the population mean + deflections is a truer interpretation of what that model notation really means? I buy that.</p><p> </p><p>When I was writing this update I was actually thinking about adding an admonition block saying that you can get an equivalent model with the population + deflection approach. But now I am swayed to go with your population + deflection approach AND update the pymc model accordingly. But I might still add in an admonition block describing this initial model, and to link to the Bambi discussion.</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/697/discussion/#comment-2323359808'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

tomicapretto:
<p>Yes, that's correct! The Wilkinson's model formula was adapted in a non-bayesian setting where you have fixed and random parameters. The population mean is the fixed part, the group deflections are the random part. In a bayesian context they're not different (both are random, and that's fine).</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/697/discussion/#comment-2323359808'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
@tomicapretto So this is all very interesting and I feel like I'm learning a lot.

I've written the Model3 (`1 + x + (1 + x | g)`) as:

Screenshot 2024-09-04 at 10 14 03

**Question 1:** If you want to describe the slope of an as yet unobserved group, you would take samples from $\text{Normal}(\beta_1, p_{1\sigma})$? And you'd do that on a sample by sample basis, i.e. that the $\beta_1$ and $p_{1\sigma}$ represent samples drawn from the posterior. It feels a but clumsy, so I just wanted to check my thinking with you.

**Question 2:** What would determine if you should specify the distributions of $u$ as $\text{ZeroSumNormal}$, rather than $\text{Normal}$?

tomicapretto:
@drbenvincent 

> Question 1: If you want to describe the slope of an as yet unobserved group, you would take samples from Normal(β1, p_1)?  And you'd do that on a sample by sample basis, i.e. that the β1 and p_1 represent samples drawn from the posterior. It feels a but clumsy, so I just wanted to check my thinking with you.

It's actually a bit more complicated than that. I think I should write a comprehensive guide for this in the future (which would help me to understand it better too). I always point people to these two links, which I find useful.

https://www.andrewheiss.com/blog/2021/11/10/ame-bayes-re-guide/
https://x.com/tjmahr/status/1458443667748818955

> Question 2: What would determine if you should specify the distributions of u as ZeroSumNormal, rather than Normal?

I'm not sure if you're asking what to do to predict for a new group with ZSN or when one should use ZSN instead of a simple Normal. As usual, it depends. 

Hopefully the simpler exampe below will help. 
Imagine you have a continuous outcome in the reals and A groups. 
We use a normal observational model like the following

$$
\begin{aligned}
Y_i &\sim \text{Normal}(\mu_i, \sigma^2) \\ 
\mu_i &= \alpha + \beta_{j[i]} \\
\alpha &\sim \text{Normal} \\
\beta_j &\sim \text{Normal} \quad \forall j = 1, \dots, A \\
\sigma &\sim \text{Something}
\end{aligned}
$$

This is a model where we attempt to have a global mean $\alpha$ and a group specific parameter $\beta_j$.
The posterior means of $\alpha$ and $\beta_j$ are non-identifiable as they're linearly dependent.

There are several things one can do to resolve that issue:

1. Constrain $\alpha$ to zero, which one think of as removing $\alpha$ from the model. In this case, the $\beta_j$ parameters represent the group means.
2. Constrain one $\beta_j$ to zero, which is like removing one $\beta_j$. $\alpha$ becomes the mean of the group for which we removed/constrained the $\beta$ parameter. All the other $\beta_j$ are deflections between the reference group mean and the $j$-th group mean.
3.  Constraint the sum of $\beta_j$ params to zero. $\alpha$ represents the global mean and each $\beta_j$ deflections around that mean.

With any of the choices above we get **equivalent** models, but using different parametrizations. All of them are "no pooling" models.

Another way to "resolve" that issue is to use a hierarchical approach. One could write the model this way

$$
\begin{aligned}
Y_i &\sim \text{Normal}(\mu_i, \sigma^2) \\ 
\mu_i &= \alpha + \beta_{j[i]} \\
\alpha &\sim \text{Normal} \\
\beta_j &\sim \text{Normal}(0, s^2) \quad \forall j = 1, \dots, A \\
s &\sim \text{Something} \\
\sigma &\sim \text{Something}
\end{aligned}
$$

or this other way

$$
\begin{aligned}
Y_i &\sim \text{Normal}(\mu_i, \sigma^2) \\ 
\mu_i &= \beta_{j[i]} \\
\beta_j &\sim \text{Normal}(m, s^2) \quad \forall j = 1, \dots, A \\
m &\sim \text{Normal} \\
s &\sim \text{Something} \\
\sigma &\sim \text{Something}
\end{aligned}
$$

where $m$ is the $\alpha$ from above.

With this hierarchical approach _we are not_ removing linear dependencies like above. What happens here is a bit more sophisticated. The $\beta_j$ parameters are no longer independent because they're assumed to be draws of a normal population with a common mean ($m$) and standard deviation ($s$). But there's no "hard" constraint like when we set some parameter, or their sum, to zero. Think of it as a "soft" constrained. 

However, as the sample size for each group $j$ increases, the posterior will converge to the posterior one would get in the first model we got above, without restrictions, which has non-identifiable means. 

What do I want to say with all of this? That, theoretically, the hierarchy will make your parameters identifiable because they're no longer independent and thus Normal should work. However, as the sample sizes for the groups grow, it will be harder and harder to get draws from the posterior because of the increasing correlation. 
If you constrain the partially-pooled $\beta_j$ parameters with a zero-sum constraint, that problem should disappear. However, that can create other issues (for example, how to get predictions for new groups). Adrian and Luciano solved that for some cases, but I'm not sure if they solved it for all cases.


See, for example, the debate here about zero-sum constraints https://discourse.mc-stan.org/t/new-stan-data-type-zero-sum-vector/26215/3.

It's a very interesting and challenging topic.

Finally, there's also this book https://www.amazon.com/Richly-Parameterized-Chapman-Statistical-Science/dp/0367533731 which covers some of this topics (identifiability) using the "degrees of freedom" concept. Long story short, when you have partially pooled parameters the degrees of freedom are a non integer value. For the model above, it should be between those of the identifiable no-pooling model and the non-identifiable partial-pooling model. 

Btw, I'm fairly familiar with this but I'm certainly not a truly expert, so I recommend reading the book.

drbenvincent:
Superb @tomicapretto !

Because of time constraints and considering the focus of the notebook, I think I will not tackle the population mean (or similar) issue in this notebook. I will take it as a win that this notebook started with the pymc model code, but now makes a reasonable attempt to discuss the mapping onto mathematical and Wilkinson descriptions.

Nevertheless I'll continue to learn about this in more and either return to give more depth to this notebook or come up with an alternative. But in the mean time it sounds like you may well be coming up with your own Capretto definitive resource blog post or docs :)

tomicapretto:
> Superb @tomicapretto !
> 
> Because of time constraints and considering the focus of the notebook, I think I will not tackle the population mean (or similar) issue in this notebook. I will take it as a win that this notebook started with the pymc model code, but now makes a reasonable attempt to discuss the mapping onto mathematical and Wilkinson descriptions.
> 
> Nevertheless I'll continue to learn about this in more and either return to give more depth to this notebook or come up with an alternative. But in the mean time it sounds like you may well be coming up with your own Capretto definitive resource blog post or docs :)

Haha thanks :D

tomicapretto:
<p>We then show 2 "wayes" to resolve... typo?</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/697/discussion/#comment-2323307556'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Done :)</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/697/discussion/#comment-2338875061'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Well spotted. Fixed</p>

---

 <a href='https://app.reviewnb.com/pymc-devs/pymc-examples/pull/697/discussion/#comment-2323307556'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
Thanks @tomicapretto. I've made those changes.

tomicapretto (approved):

juanitorduz (approved):

PR #703, opened 2024-09-19
Fix img path issues

B-Deforce:
The following images `pooled.png`, `unpooled.png`, and `partial_pooled_model.png` were wrongly assumed to reside in the same folder as the notebook and referred to as such in the notebook. Instead, they are located in `../case_studies/`. These paths have now been fixed.

----
📚 Documentation preview 📚: https://pymc-examples--703.org.readthedocs.build/en/703/


OriolAbril:
Thanks! Can you instead move the images to the generalized linear models folder? The notebook used to be in the case studies folder: https://github.com/pymc-devs/pymc-examples/tree/bed70e39355b6f4286b1c7aafa337e23bffa612b/examples/case_studies and it looks like it was moved without the images. Now that you have pointed out the issue, I don't think it makes sense to continue keeping the images in a different folder

B-Deforce:
You're right, moving the images makes more sense. Fixed it :).

juanitorduz (approved):

PR #704, opened 2024-09-19
improve prior comment on GP Birthdays data set.

juanitorduz:
The comment in https://www.pymc.io/projects/examples/en/latest/gaussian_processes/GP-Births.html#prior-specifications is confusing as the log transform is tasked on the y variable and has nothing to do with the time dimension (as noted by @AlexAndorra )

----
📚 Documentation preview 📚: https://pymc-examples--704.org.readthedocs.build/en/704/


fonnesbeck (approved):
LGTM

Issue #735, opened 2024-11-15
Error in `examples/spatial/malaria_prevalence.ipynb`


wd60622:
Something up with the CRS using these versions:

```python
rasterio.__version__
# '1.4.2'
pyproj.__version__
# '3.7.0'
mapclassify.__version__
# '2.8.1'
```

Here is the tracback:

```python
──────────────────────────────────────────── Error running examples/spatial/malaria_prevalence.ipynb ─────────────────────────────────────────────

---------------------------------------------------------------------------
Exception encountered at "In [7]":
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[7], line 13
      9 with rio.open(in_path) as src:
     11     img = src.read()
---> 13     src_crs = src.crs["init"].upper()
     14     min_lon, min_lat, max_lon, max_lat = src.bounds
     15     xs = gambia_gpdf_4326["geometry"].x

File rasterio/crs.pyx:885, in rasterio.crs.CRS.__getitem__()

KeyError: 'init'
```

PR #745, opened 2024-11-17
fixed bug introduced by rasterio update

Dekermanjian:
# Fixed bug introduced by `rasterio` update
Resolves issue #735 

----
📚 Documentation preview 📚: https://pymc-examples--745.org.readthedocs.build/en/745/


OriolAbril (commented):
Fix looks good. I just realized that the notebook uses the extra dependencies template but doesn't specify which are the dependencies in the metadata. Do you mind fixing this here too?

Here is how the dropdown looks for this notebook:

imatge


And how it looks for one of the HSGP ones that also uses the template:

imatge

Dekermanjian:
@OriolAbril yes, I will definitely fix it! Sorry, I didn't realize it wasn't correct earlier.

Dekermanjian:
Hey @OriolAbril, how do I add the dependencies into the metadata? I tried adding it straight into the associated .myst.md file but that doesn't seem the right way to do it.

OriolAbril:
changes in the myst files get overriden by the contents of the jupyternotebook in pre-commit. You have to either follow the instructions in https://www.pymc.io/projects/docs/en/latest/contributing/jupyter_style.html#extra-dependencies or modify the myst file, then manually call `jupytext` to convert myst->ipynb before pre-commit (not sure if outputs will be removed when doing this so I'd recommend the first option)

OriolAbril:
I don't use vscode but it would be nice to add how to do it from there to the docs

OriolAbril (approved):

Issue #747, opened 2024-11-21
Use pm.Censored in accelerated failure model


ricardoV94:
**Notebook title**:
**Notebook url**:

Instead of using a Potential use Censored

ricardoV94:
> Hey @ricardoV94 would it be alright if I worked on this?

Yes, you can claim any open issue by opening a PR. Until then it's up for grabs

Dekermanjian:
Understood! Thank you, @ricardoV94 !

PR #750, opened 2024-11-25
replaced all pymc potential with pymc censored

Dekermanjian:
# Update AFT Weibull notebook to use `pm.Censored` instead of `pm.Potential`
Related to #747 

I updated all 3 parameterizations of the AFT Weibull model to utilize `pm.Censored` instead of `pm.Potential`. I also added into the markdown, after the `pm.Potential` is discussed, why we would rather use `pm.Censored`.

For the 1st two parameterizations the estimates are the same using `pm.Censored` as they are with `pm.Potential` for the last parameterization the parameter estimates are slightly shifted.


fonnesbeck (approved):
Thanks for this!

Dekermanjian:
@fonnesbeck 
Absolutely! Happy to contribute!

Issue #743, opened 2024-11-16
Error in `examples/case_studies/probabilistic_matrix_factorization.ipynb`


williambdean:
Error while finding MAP for the model.

```python
────────────────────────────────── Error running examples/case_studies/probabilistic_matrix_factorization.ipynb ──────────────────────────────────

---------------------------------------------------------------------------
Exception encountered at "In [20]":
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
Cell In[20], line 2
      1 # Find MAP for PMF.
----> 2 pmf.find_map();

Cell In[13], line 6, in _find_map(self)
      4 with self.model:
      5     logging.info("finding PMF MAP using L-BFGS-B optimization...")
----> 6     self._map = pm.find_MAP(method="L-BFGS-B")
      8 elapsed = int(time.time() - tstart)
      9 logging.info("found PMF MAP in %d seconds" % elapsed)

File ~/micromamba/envs/pymc-examples/lib/python3.11/site-packages/pymc/tuning/starting.py:134, in find_MAP(start, vars, method, return_raw,
include_transformed, progressbar, progressbar_theme, maxeval, model, seed, *args, **kwargs)
    127 ipfn = make_initial_point_fn(
    128     model=model,
    129     jitter_rvs=set(),
    130     return_transformed=True,
    131     overrides=start,
    132 )
    133 start = ipfn(seed)
--> 134 model.check_start_vals(start)
    136 vars_dict = {var.name: var for var in vars}
    137 x0 = DictToArrayBijection.map(
    138     {var_name: value for var_name, value in start.items() if var_name in vars_dict}
    139 )

File ~/micromamba/envs/pymc-examples/lib/python3.11/site-packages/pymc/model/core.py:1742, in Model.check_start_vals(self, start, **kwargs)
   1736     valid_keys = ", ".join(value_names_set)
   1737     raise KeyError(
   1738         "Some start parameters do not appear in the model!\n"
   1739         f"Valid keys are: {valid_keys}, but {extra_keys} was supplied"
   1740     )
-> 1742 initial_eval = self.point_logps(point=elem, **kwargs)
   1744 if not all(np.isfinite(v) for v in initial_eval.values()):
   1745     raise SamplingError(
   1746         "Initial evaluation of model at starting point failed!\n"
   1747         f"Starting values:\n{elem}\n\n"
   1748         f"Logp initial evaluation results:\n{initial_eval}\n"
[...]

PR #834, opened 2026-01-18
Update PMF example

ckrapu:
# Update Probabilistic Matrix Factorization example

Address #743 and #58, updating this notebook to run cleanly and also performing some light edits for clarity and correctness.

+ [x] PR description contains a link to the relevant issue:

### Helpful links


ckrapu:
The only substantial change in the model for this notebook was a replacement of the `MvNormal` RVs which had a diagonal covariance matrix with independent draws from `Normal` instead.

One odd thing that I found was that the MAP optimization finished in <1 second on the old notebook, but took 40 seconds this time. Also, the performance difference between the mean-of-means baseline and the MAP estimate was not reproducible, but the answer we get now (MAP is closer to baseline performance instead of much worse) makes more sense to me. I also added a few bullet points to explain the dims of the various matrices.

aloctavodia:
Thanks @ckrapu!

PR #840, opened 2026-02-09
Added Probabilistic PCA example

### LINKED ISSUE #250: Add Probabilistic PCA example (code already exists)
I coded this up a while ago, pretty simple model but I think it's an interesting (and commonly referred to) model: https://gist.github.com/twiecki/713cf58f5fa37332d0cda48a7c798016

Probably just needs some more text and explanation.

ParthsarthiSingh-glang:
## Summary
Adds a comprehensive Probabilistic PCA example to the case studies section.

Closes #250

## Contents
- Mathematical background on PPCA
- Full PyMC implementation with priors
- Comparison with classical PCA (sklearn)
- Convergence diagnostics and trace plots
- Discussion of rotational identifiability

## Testing
- [x] Notebook runs successfully end-to-end
- [x] All cells execute without errors
- [x] Visualizations render correctly

## Notes
- ESS warnings are expected due to rotational non-identifiability
- Tested on Windows with Python 3.13


ParthsarthiSingh-glang:
@twiecki Thanks for taking a look!

You're right - the convergence warnings are there because of the rotational symmetry in PPCA. Since you can rotate the latent space and get the same answer, the sampler has trouble settling down on one specific solution.

But the model is working correctly - the reconstructions look good and match what sklearn's PCA produces. It's just that the diagnostics don't look great because of this rotation issue.

Would it help if :
- Increase the sampling to get better numbers?
- Add a note explaining why this happens with PPCA?

I can adjust on whatever you tell .

twiecki:
Yes, I'd add a note explaining that.

Issue #862, opened 2026-04-23
Updates examples for new major releases of pymc/pytensor/arviz


jessegrabowski:
Arviz v1 is going to break everything for sure, but there might be small details that will stop working due to pymc v6/pytensor v3. This issue is a tracker for checking that everything works post-release.

*Introductory*
- [ ] pymc_overview
- [ ] glm_linear
- [x] api_quickstart

*Library Fundamentals*
- [ ] dimensionality
- [ ] pymc_pytensor
- [x] data_container

*How To*
- [ ] posterior_predictive
- [ ] model_comparison
- [ ] lkj
- [ ] missing_data_imputation
- [ ] blackbox_external_likelihood_numpy
- [ ] copula_estimation
- [ ] debugging
- [x] hypothesis_testing
- [ ] marginalizing_models
- [ ] model_builder
- [ ] profiling
- [ ] spline
- [ ] updating_priors
- [ ] wrapping_jax_function

*Generalized Linear Models*
- [x] glm_binomial_regression
- [x] glm_discrete_choice_models
- [x] glm_hierarchical_binomial_model
- [ ] glm_missing_values_in_covariates
- [ ] glm_model_selection
- [ ] glm_negative_binomial_regression
- [ ] glm_ordinal_features
- [ ] glm_ordinal_regression
- [ ] glm_out_of_sample_predictions
- [ ] glm_poisson_regression
- [ ] glm_robust_with_outlier_detection
- [ ] glm_robust
- [ ] glm_rolling_regression
- [ ] glm_truncated_censored_regression
- [ ] multilevel_modeling

*Case Studies*
- [x] best
- [ ] cfa_sem
- [x] gev
- [ ] bayesian_sem_workflow
- [x] bayesian_workflow
- [x] binning
- [x] factor_analysis
- [x] hierarchical_partial_pooling
- [x] item_response_nba
- [x] occupancy
- [ ] probabilistic_matrix_factorization
- [x] putting_workflow
- [x] reinforcement_learning
- [ ] reliability_and_calibrated_prediction
- [x] rugby_analytics
- [ ] ssm_hurricane_tracking

*Causal Inference*
- [ ] glm_simpsons_paradox
- [ ] bayesian_ab_testing_introduction
- [ ] bayesian_nonparametric_causal
- [x] counterfactuals_do_operator
- [ ] difference_in_differences
- [ ] excess_deaths
- [ ] interrupted_time_series
- [ ] interventional_distribution
[...]

PR #882, opened 2026-05-24
update variational_api_quickstart for PyMC 6

YichengYang-Ethan:
## Summary
- Wrap `Approximation.sample(...)` calls in `with <model>:`; PyMC 6 requires an active model context (cells 9, 21, 50, 53). Part of #862.
- Replace removed `az.plot_posterior` with `az.plot_dist(idata, var_names=[...])` (cell 21).
- Move `pm.callbacks.{Tracker,CheckParametersConvergence}` references to `pm.variational.callbacks` (cells 28, 40, 84).
- Switch `total_size=data.shape` (tuple) to `data.shape[0]` (int) — the tuple form hits `IndexError` in `get_scaling` (`minibatch_rv.py:96`) (cell 100).

Verified with Restart & Run All under pymc 6.0.1 / pytensor 3.0.3 / arviz 1.1.0; pre-commit passes.


YichengYang-Ethan:
@aloctavodia thanks for the thorough review! Addressed all eight comments in the new commit:

- `arviz-variat` style applied at the top, no more `%matplotlib inline`
- `ipywidgets` installed locally; widget warnings cleared from outputs
- Model comparisons (cells 10, 50, 53) now use `az.plot_dist({"name": idata, ...})` dict form
- `az.plot_trace` → `az.plot_trace_dist` (kept it over `plot_rank_dist` to stay consistent with the existing "traces for x² and sin(x)" narrative — happy to switch if you'd prefer rank plots here)
- Cells 61, 64: `az.convert_to_dataset` + `az.plot_dist`; the first one also switches to `sample_node(size=N)` so the plot reflects independent samples
- Cell 60 markdown: noted that repeated `.eval()` on a `sample_node` returns the cached value in modern PyTensor, and pointed to `sample_node(size=N)` for fresh samples

Ready for re-review.

YichengYang-Ethan:
Thanks @aloctavodia, applied both:

- `pc.add_legend("model")` + `visuals` dict to keep only the KDE in cell 10
- semicolon on `advi.approx` in cell 34

Applied the same legend + visuals pattern to the two other multi-model `plot_dist` calls (cells 50 and 53) so the look stays consistent — happy to revert if you'd rather only have it on the first one.

aloctavodia (approved):

PR #888, opened 2026-06-09
Example: out-of-core minibatch variational inference with DataLoader and Trainer

YichengYang-Ethan:
Draft for mentor review (GSoC 2026, Streaming Variational Inference). Pairs with pymc-devs/pymc#8325. Not for merge yet.

A how-to notebook for out-of-core minibatch variational inference with the new `pymc.variational.streaming` API: a `DataLoader` over `parquet_source`, a `pm.Data` placeholder, and the callback-free `Trainer`.

It writes a synthetic logistic dataset to Parquet shards, deletes the combined in-memory table (`X` and `y` stay only for the in-RAM baseline comparison), streams minibatches off disk with ADVI, and shows that the streaming posterior agrees with an in-RAM `pm.Minibatch` fit on the toy problem while peak memory stays flat in N. Both fits and both posterior draws are seeded, so the comparison is reproducible.

The memory-at-scale numbers come from a separate run of the same model on the public Criteo 1TB benchmark. The notebook itself runs end to end in seconds, and the outputs and figures are committed.

Open question: where the example data should live for a fully self-contained run (a download script vs HuggingFace hosting). Happy to raise this in Discord.


YichengYang-Ethan:
Closing this in favour of the proposal in #891 and the replacement PR #892: this draft was written against the pre-merge DataLoader API (`sample_shape=`, `len(loader)` as the row count), both of which changed when pymc-devs/pymc-extras#698 merged, and the replacement notebook covers the same ground end to end on a self-contained dataset. One example, not two competing ones.

Issue #891, opened 2026-08-14
Out-of-core minibatch ADVI on a financial tick stream


YichengYang-Ethan:
# Notebook proposal
**Title**: Out-of-core minibatch ADVI on a financial tick stream

## Why should this notebook be added to pymc-examples?

Once a dataset outgrows memory, PyMC users currently have no worked example to follow: every VI and minibatch example assumes the data sits in RAM. The DataLoader merged in pymc-devs/pymc-extras#698 removes that limit, and this notebook is its worked example, end to end on realistic financial tick data (fully self-contained: the data is generated inside the notebook, no downloads).

It demonstrates the parts that go wrong silently if skipped, which is what makes it a tutorial rather than an API tour: why the likelihood must factor over rows for minibatch rescaling to be valid (and why the classic stochastic-volatility model therefore cannot be minibatched — this notebook is the complement to that example, not a replacement); a disk-level shuffle for strongly ordered data, since a bounded buffer only block-shuffles; the `total_size` contract and what the mismatch looks like; a hurdle likelihood for the 30% of ticks with exactly zero price move; and loss-based stopping on a noisy streamed ELBO, including how a naive per-step rule false-stops.

Suggested categories:
* Level: intermediate
* [Diataxis](https://diataxis.fr/) type: How-to

## Related notebooks

- `variational_api_quickstart` — this builds directly on it, adding the out-of-core data path.
- `stochastic_volatility` — cited as the contrast case: a latent-path likelihood that does not factor over rows and so cannot be minibatched; this notebook explains why and shows a model class that can.
- GLM hierarchical / minibatch examples — same `total_size` machinery, in-memory; this notebook moves it out of core.
- Supersedes the draft in pymc-examples#888, which was written against the pre-merge DataLoader API; that PR will be closed pointing here so there is one example, not two.

# References

- pymc-devs/pymc-extras#698 (the merged DataLoader), #733 (loss-based convergence callback, in review).
- Kucukelbir et al. (2015), Automatic Differentiation Variational Inference — the `total_size` minibatch rescaling.
- Page (1954), Continuous inspection schemes — the change-detection framing used in the stopping section.

Note on dependencies: the DataLoader is on pymc-extras main but not yet in a released version (latest is v0.14.0, 2026-07-28); the notebook targets the next release.

PR #889, opened 2026-07-07
Example discrete markov chain weather

juanitorduz:
port of https://github.com/pymc-devs/pymc-extras/pull/702

PR #892, opened 2026-08-14
Out-of-core minibatch ADVI on a financial tick stream

YichengYang-Ethan:
Proposal issue: #891. Supersedes #888, which was written against the pre-merge DataLoader API.

An end-to-end example of out-of-core minibatch ADVI: 300,000 rows of synthetic tick data generated inside the notebook (no downloads), streamed from Parquet through the `DataLoader` merged in pymc-devs/pymc-extras#698 into a hierarchical hurdle Bernoulli + Student-t model with `total_size` rescaling. `DataLoader` lives on pymc-extras `main` and is not in a release yet (latest is v0.14.0); the notebook carries a pinned install note (`pip install git+https://github.com/pymc-devs/pymc-extras@8db1880`) until it is.

**Files.** The notebook pair, and `examples/references.bib`: seven entries added, all cited here (Hoffman et al. 2013, Engle–Russell 1998, Andersen–Bollerslev 1997, Bollerslev 1987, Roll 1984, Page 1954, Polyak–Juditsky 1992). Nothing else in the bib is touched; the branch is rebased on current `main` (#889).

**What it covers, in reading order.**

- Two acceptance gates for minibatch VI — the likelihood must factor over rows, and there must be no low-dimensional sufficient statistic — with the stochastic-volatility notebook as the counter-example, and a model class that passes both.
- The model on the event clock, with every symbol defined under the equation.
- A seeded generator with each setting annotated; a disk-level hash shuffle and equal 1,000-row row groups so the `shuffle=False` path yields exactly 300 blocks per epoch, counted rather than inferred.
- The streaming fit through a twenty-line `pm.fit` callback; why `approx.hist` is on the batch scale (MinibatchRV scales the observed term by N/b, OPVI divides the objective back) and the N/b rescaling that puts it on the full-data negative-ELBO scale.
- Why the last iterate is not the answer: with a fixed replay order the iterate carries an order-locked component (median swing 4.3 posterior sd within an epoch, against 0.11 sd between the same phase of consecutive epochs), so the variational parameters are averaged over the final pass before anything is read off.
- Recovery restricted to identified combinations, with a z column; the ridge rows explained (flat likelihood, hierarchical prior tilting toward zero-mean effects, mean-field reporting the conditional width across the ridge rather than the width along it); a warm-start continuation showing those coordinates still crawling.
- Replication: the nine identified rows read four ways, each fit judged in its own posterior width — last iterate 3.42, tail-averaged 1.16, a different optimizer seed on the same order 1.14, a different on-disk order 1.13 — with the notebook explicit that this vetoes instability and does not certify calibration.
- Symbol-effect contrasts against the generating values, the two event-clock curves for one symbol, and an in-sample posterior predictive check on the zero share and the tail.
- What a stopping rule has to be able to see: standardized block contrasts on this trace at per-step, one-epoch and two-epoch horizons plus a 1.5-epoch control, and why a per-step CUSUM fires instead of staying silent.
- A closing "Related tooling" table naming the four pymc-extras pieces (pymc-devs/pymc-extras#698 merged; pymc-devs/pymc-extras#710, pymc-devs/pymc-extras#722, pymc-devs/pymc-extras#733 open), and a pre-fit applicability note on why streaming Pathfinder is not run on a 154-parameter hierarchical target.

**Deliberate restrictions.** Only merged library code is imported; nothing is asserted that the notebook does not compute; every statement about what the fit recovers is conditional on the generator; no claim is made about width calibration.

**Provenance.** Stored outputs come from a clean environment at pymc-extras `8db1880`, PyMC 6.2.0, PyTensor 3.2.4 — what the watermark reports and what the install note tells readers to use. The notebook times itself (42 s in the stored run on a loaded 10-core arm64 laptop, about 28 s idle; three fits plus a sixty-pass continuation), peak RSS about 640 MB. Pre-commit passes and the docs preview builds.

**Style.** Checked against the Jupyter style guide and the five `variational_inference` siblings. Two deliberate departures a reviewer may notice: the prose is impersonal rather than "we", and the figures are hand-rolled matplotlib rather than ArviZ calls, each for a reason stated in its cell comment (truth overlays, the annotated zero jump, the two-panel horizon view). Happy to switch either if you prefer the section's convention.

**One question for a maintainer.** `pixi.toml` pins `pymc-extras >=0.5,<0.6` and does not list `pyarrow`, so the repo environment cannot execute this notebook. `nb_execution_mode` is `off`, so nothing in the build depends on that, and the install note gives the exact command that reproduces the stored outputs. I have not touched `pixi.toml` because widening it is a repo-wide call rather than mine — tell me if you would rather this PR carried that change.

AI disclosure: I used Claude extensively on this notebook, for drafting, the ETL harness, and re-verifying the numeric claims; every number was checked by running code, and the prose was reviewed and edited by hand.


YichengYang-Ethan:
Trimmed to the bare minimum ahead of review, per the menu in the description: the full-corpus results section is cut entirely (the notebook is now fully self-contained, ~1 minute end to end), the design narrative is tightened, and an install-from-main note covers the release gap. The stopping-rule section stays; it becomes a short reference to pymc-devs/pymc-extras#733 once that merges. 1416 lines down to 941, re-executed, pre-commit green.

PR #640, opened 2024-02-26
sitemap fix

OriolAbril:
- try fixing sitemap urls
- update robots.txt
- fix url template for sitemap entries

----
📚 Documentation preview 📚: https://pymc-examples--640.org.readthedocs.build/en/640/

PR #639, opened 2024-02-26
Add website metadata

OriolAbril:
This adds the sphinx-sitemap extension and a robots.txt file.

I have followed the adivse in https://sphinx-sitemap.readthedocs.io/en/latest/search-optimization.html but I don't really know what I'm doing, someone should really check the contents of both robots.txt and sitemap.xml: https://pymcio--639.org.readthedocs.build/projects/examples/en/639/sitemap.xml

----
📚 Documentation preview 📚: https://pymc-examples--639.org.readthedocs.build/en/639/

PR #893, opened 2026-08-31
Fix broken canonical URLs in the example gallery

twiecki:
## Bug

`html_baseurl` is set to `https://www.pymc.io/projects/examples/`, and Sphinx builds each page's canonical URL as `html_baseurl + pagename` — producing URLs without the `/en/latest/` segment that redirect to a 404. So every example notebook page ships a broken canonical tag.

Search impact (16-month Google Search Console export for pymc.io): the example gallery is the site's biggest section (~5M impressions, 66k clicks), old dated builds like `/en/2021.11.0/` still collect hundreds of thousands of impressions because nothing consolidates them onto the current pages, and GSC reports thousands of 404s.

## Fix

- `html_baseurl` now includes `/en/latest/` (the default version), so canonicals point at real pages. Sitemap URLs are unchanged — the version prefix moves from `sitemap_url_scheme` into the base URL (verified against the live sitemap format).
- `blog_baseurl` moves off the retired `docs.pymc.io` domain to `www.pymc.io` (affects feed URLs).

Companion PRs: pymc-devs/pymc.io#124 (homepage + robots.txt, which also blocks the outdated dated builds from crawling) and pymc-devs/pymc#8418 (same canonical bug in the docs).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/conf.py
+++ b/examples/conf.py
@@ -71,9 +71,11 @@ def setup(app: Sphinx):
 
 # theme options
 html_theme = "pymc_sphinx_theme"
-html_baseurl = "https://www.pymc.io/projects/examples/"
+# Also used to build each page's canonical URL, so it points at the default
+# (latest) version: <html_baseurl> + <page name>.html
+html_baseurl = "https://www.pymc.io/projects/examples/en/latest/"
 rtd_version = os.environ.get("READTHEDOCS_VERSION", "")
-sitemap_url_scheme = f"{{lang}}{rtd_version}/{{link}}"
+sitemap_url_scheme = "{link}"
 html_theme_options = {
     "secondary_sidebar_items": ["postcard", "page-toc", "edit-this-page", "sourcelink", "donate"],
     "navbar_start": ["navbar-logo"],
@@ -127,7 +129,7 @@ html_sidebars = {
 }
 
 # ablog config
-blog_baseurl = "https://docs.pymc.io/projects/examples/en/latest/"
+blog_baseurl = "https://www.pymc.io/projects/examples/en/latest/"
 blog_title = "PyMC Examples"
 blog_path = "blog"
 blog_authors = {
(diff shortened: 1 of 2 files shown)


twiecki:
Added a second commit: the generated gallery page now gets an explicit meta description and a one-sentence intro — its search snippet was previously built from concatenated card titles (269k impressions at 1.3% CTR). Verified the multiline `.. meta::` directive renders correctly with a standalone Sphinx build.

twiecki:
Could someone take a look at this example-gallery canonical URL fix? It has been open for a little over a week without review; the checks and docs preview passed. Happy to address feedback.

fonnesbeck:
### Code review

Found 1 issue:

1. Sitemap URLs for non-`latest` Read the Docs builds now all point at `/en/latest/`. `sphinx-sitemap` composes each `<loc>` as `html_baseurl + sitemap_url_scheme`, so hard-coding `en/latest/` into `html_baseurl` and dropping `rtd_version` from `sitemap_url_scheme` means the `stable` and tagged builds (e.g. `2026.02.0`, both active on RTD) emit a `sitemap.xml` that describes `latest` rather than themselves. The PR description says "Sitemap URLs are unchanged", but that only holds for the `latest` build. This undoes the per-version behavior deliberately introduced in #640 (follow-up to the "sitemap doesn't seem to work" report in #639). The canonical-tag change is fine; if collapsing every version's sitemap onto `latest` is intended, a comment next to `sitemap_url_scheme` saying so would prevent a future "fix". Otherwise keep the sitemap scheme version-aware (e.g. `sitemap_url_scheme = f"{{lang}}{rtd_version}/{{link}}"` against a versionless base) while pinning only the canonical URL to `latest`.

https://github.com/pymc-devs/pymc-examples/blob/fc0a2b92046a9218c9f6af18b5cb6de5e95a41ff/examples/conf.py#L74-L79

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

fonnesbeck (approved):
LGTM

